PR #1, opened 2023-09-27
Support nvmrc

ericanderson:
Can someone verify if this works? I'm shooting in the dark.


pashakondratyev (approved):
`nvm use` picks up this version if that's the intention

PR #10, opened 2023-10-02
Add generated types, errors, and requests for the OpenApi definition.

pashakondratyev:
Adds all the generated types from the OpenApi definition. In the future we may want this likely to be done as some build job.


packages/api/src/client/internal/generated/openapi/namespaces/Datasets.ts
```diff
@@ -0,0 +1,591 @@
...
+ * Third-party applications using this endpoint via OAuth2 must request the following operation scope: `api:datasets-write`.
+ *
+ */
+export function createDataset<TResponse>(_request: OpenApiRequest<Dataset, TResponse>, request: CreateDatasetRequest): Promise<TResponse> {
+    return _request(
+        "createDataset",
+        "POST",
+        "/v1/datasets",
+        request,
+        __undefined,
```
ericanderson, line 33: You can just generate `,` instead of `__undefined,`.

packages/api/src/client/internal/generated/openapi/request/OpenApiRequest.ts
```diff
@@ -0,0 +1,11 @@
+export type OpenApiRequest<TExpectedResponse, TResponse = TExpectedResponse> = (
+    endpointName: string,
+    method: string,
+    endpointPath: string,
+    data?: any,
+    headers?: { [header: string]: string | number | boolean | undefined | null; },
```
ericanderson, line 6: I think custom headers are pretty rare right? Certainly more rare than queryArguments. Given arguments can be omitted in generated code, we can slim things down a bit by ordering these by frequency of use.

packages/api/src/client/internal/generated/openapi/request/OpenApiRequest.ts
```diff
@@ -0,0 +1,11 @@
+export type OpenApiRequest<TExpectedResponse, TResponse = TExpectedResponse> = (
```
ericanderson, line 1: Is there an implementation of this somewhere?

packages/api/src/client/internal/generated/openapi/namespaces/OntologiesV2.ts
```diff
@@ -0,0 +1,971 @@
...
+ * Gets a specific ontology with the given Ontology RID.
+ *
+ * Third-party applications using this endpoint via OAuth2 must request the following operation scope: `api:read-data`.
+ *
+ */
+export function getOntologyV2<TResponse>(_request: OpenApiRequest<OntologyV2, TResponse>, ontology: OntologyApiName): Promise<TResponse> {
+    return _request(
+        "getOntologyV2",
+        "GET",
+        "/v2/ontologies/{ontology}",
```
ericanderson, line 85: Given we have template literals, does it make sense to have a special section for passing along query params vs just putting them in place?

packages/api/src/client/internal/generated/openapi/namespaces/Operations.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function getOperation<TResponse>(_request: OpenApiRequest<AsyncOperation, TResponse>, operationId: string, preview?: PreviewMode): Promise<TResponse> {
+    return _request(
+        "getOperation",
+        "GET",
+        "/v2/operations/{operationId}",
+        __undefined,
+        __undefined,
+        {
+            preview,
```
ericanderson, line 19: What is the rationale between taking `operationId: string, preview?: PreviewMode` vs taking `operationId: string, customHeaders?: { preview?: PreviewMode }`? Basically, why make things more complex than they need to be?

packages/api/src/client/internal/generated/openapi/namespaces/OntologiesV2.ts
```diff
@@ -0,0 +1,971 @@
...
+        "/v2/ontologies/{ontology}/models/deployments/{deployment}",
+        __undefined,
+        __undefined,
+        __undefined,
+        {
+            ontology,
+            deployment,
+        },
+        __undefined,
+        "application/json",
```
ericanderson, line 942: Given 99% of our APIs all expect json in return and if they have a body, send JSON, can we just make these the defaults for our custom request function? This lets these be omitted and slim down the code quite a bit.

packages/api/src/client/internal/generated/openapi/components/ActionParameterType.ts
```diff
@@ -0,0 +1,267 @@
...
+import { DateType } from "./DateType";
+import { DoubleType } from "./DoubleType";
+import { IntegerType } from "./IntegerType";
+import { LongType } from "./LongType";
+import { OntologyObjectSetType } from "./OntologyObjectSetType";
+import { OntologyObjectType } from "./OntologyObjectType";
+import { StringType } from "./StringType";
+import { TimestampType } from "./TimestampType";
+
+export interface ActionParameterType_Array extends ActionParameterArrayType {
```
ericanderson, line 13: What do we get by separating the `type` out?

packages/api/src/client/internal/generated/openapi/components/Aggregation.ts
```diff
@@ -0,0 +1,152 @@
...
+    max: (obj: MaxAggregation) => T;
+    min: (obj: MinAggregation) => T;
+    avg: (obj: AvgAggregation) => T;
+    sum: (obj: SumAggregation) => T;
+    count: (obj: CountAggregation) => T;
+    approximateDistinct: (obj: ApproximateDistinctAggregation) => T;
+    unknown: (obj: Aggregation) => T;
+}
+
+function visit<T>(obj: Aggregation, visitor: AggregationVisitor<T>): T {
```
ericanderson, line 108: Im not sure we need the visitors.

ericanderson (commented):
This is awesome. I left a bunch of notes that will slim all this generated code down a lot and speed up compilation for us.

packages/api/src/client/internal/generated/openapi/components/ActionParameterType.ts
pashakondratyev: For openapi type unions we use `type` as the discriminator and that doesn't exist on the base components so if we have two components for example `A = { args: string[]}` and `B = {arg: strings}` but we want a union of the two for some request, it would require it to be like `{ type: "A", args:string[] } | {type: "B": arg: string}` so they're kind of like different interfaces

ericanderson: Yeah im more curious why we split out the underlying data from the interface that has `type: "foo"`

packages/api/src/client/internal/generated/openapi/components/Aggregation.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (isApproximateDistinct(obj)) {
+        return visitor.approximateDistinct(obj);
+    }
+
+    return visitor.unknown(obj);
+}
+
+export type Aggregation = Aggregation_Max | Aggregation_Min | Aggregation_Avg | Aggregation_Sum | Aggregation_Count | Aggregation_ApproximateDistinct;
+
+export const Aggregation = {
```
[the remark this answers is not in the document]
pashakondratyev: I think we won't need these if we're not exposing the visitors

packages/api/src/client/internal/generated/openapi/components/ActionParameterType.ts
pashakondratyev: Depending on what A and B are they can be used elsewhere so if we had the data in the same interface that has `type: "foo"` we'd need to generate some duplicate code (unless I'm misunderstanding)

packages/api/src/client/internal/generated/openapi/components/Aggregation.ts
pashakondratyev: We can remove this for now!

packages/api/src/client/internal/generated/openapi/request/OpenApiRequest.ts
pashakondratyev: I can provide an implementation, this was more just to define a contract between the generated functions and what we should pass in

packages/api/src/client/internal/generated/openapi/namespaces/Datasets.ts
ericanderson: mistaken i was

packages/api/src/client/internal/generated/openapi/namespaces/Datasets.ts
```diff
@@ -46,17 +42,11 @@ export function createDataset<TResponse>(_request: OpenApiRequest<Dataset, TResp
...
         "GET",
-        "/v1/datasets/{datasetRid}",
-        __undefined,
+        `/v1/datasets/${datasetRid}`,
         __undefined,
         __undefined,
-        {
-            datasetRid,
-        },
         __undefined,
```
ericanderson, line 49: For a future PR, we can omit all the trailing `__undefined,`

ericanderson (approved):

PR #12, opened 2023-10-02
Implement type tests for existing code

ericanderson:
Provides some sanity that we don't break the derived type APIs when we make changes.

We will need to keep adding to this as we discover new issues and implement new features.

**Bonus:** Doing this helped expose + fix properties showing up in intellisense as `adUsername: never` for aggregations instead of just omitting it. This PR also fixes that.

### DIFF
--- a/examples/basic/cli/src/examples/fetchAggregationForEmployeesGroupedThin.ts
+++ b/examples/basic/cli/src/examples/fetchAggregationForEmployeesGroupedThin.ts
@@ -17,6 +17,8 @@
 import type { ThinClient } from "@osdk/api";
 import { aggregateOrThrow } from "@osdk/api/objects";
 import invariant from "tiny-invariant";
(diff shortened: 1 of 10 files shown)


examples/basic/cli/src/examples/fetchAggregationForEmployeesGroupedThin.ts
```diff
@@ -55,7 +57,30 @@ export async function fetchAggregationForEmployeesGroupedThin(
...
-  result; //
+  expectType<
+    TypeOf<
+      Array<{
+        group: {
+          locationType: string | undefined;
+        };
+        values: {
+          "employeeNumber": {
+            max: number | undefined;
```
pashakondratyev, line 68: Should these just be number?

pashakondratyev (approved):

PR #13, opened 2023-10-03
Add turbo fix-lint command

ericanderson:


examples/basic/cli/package.json
```diff
@@ -18,6 +18,7 @@
   "scripts": {
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
     "dev:transpile": "tsup --watch",
+    "fix-lint": "eslint . --fix && dprint fmt --config $(find-up dprint.json)",
     "lint": "eslint . && dprint check  --config $(find-up dprint.json)",
     "prettier": "prettier .",
```
pashakondratyev, line 23: Should we remove `prettier` as well?

ericanderson: Good call.

examples/basic/cli/package.json
```diff
@@ -18,6 +18,7 @@
   "scripts": {
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
     "dev:transpile": "tsup --watch",
+    "fix-lint": "eslint . --fix && dprint fmt --config $(find-up dprint.json)",
```
pashakondratyev, line 21: Also is there a good convention for dashes vs colons? I.e. `fix:lint`

pashakondratyev (approved):

examples/basic/cli/package.json
ericanderson: Hmm. Not really. I'm going to leave this for now and we can come back to it later. Earlier versions of turbo didn't like the colon but its fine now obviously. 

In this case I used the colon as the dev:transpile is a "child" task of the virtual "dev" task.

ericanderson: I guess we could say that typecheck is a child of `lint` and arguably probably should be. But typescript doesnt currently let us use build mode and `--noEmit` complicating things. So it doesnt just lint check, it has a side effect of creating `build/types/*`.

Do you have a strong opinion on any of this?

PR #15, opened 2023-10-12
Using generated types for requests

pashakondratyev:
This is just a start to move the hand rolled requests and types to the generated ones from the OpenApi spec.
We can go back and remove all the `wire` imports we don't need and convert them to the generated ones. 
Also added `vitest` for testing as it has build in ESM support + TS support out of the box, required no config, and ran very fast


examples/basic/sdk/src/syntax.test.ts
```diff
@@ -178,3 +179,8 @@ async function aggregateThingsWithGroupsAndHandleErrors() {
     }
   }
 }
+
+describe("syntax", () => {
```
pashakondratyev, line 183: Empty placeholder tests so vitest doesn't complain about `no tests existing`

ericanderson (approved):

PR #19, opened 2023-10-12
Checkpoint generator progress

ericanderson:


packages/generator/tsconfig.json
```diff
@@ -10,6 +10,9 @@
     ".eslintrc.cjs"
   ],
   "references": [
+    {
```
pashakondratyev, line 13: I think we wouldn't need these if we added `types` to the package.json (I think conditional exports don't work super well and might be why you needed to add these)

pashakondratyev (approved):

packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.ts
```diff
@@ -0,0 +1,90 @@
...
+import type { WireOntologyDefinition } from "../WireOntologyDefinition";
+import { wireObjectTypeV2ToObjectDefinitionInterfaceString } from "./wireObjectTypeV2ToObjectDefinitionInterfaceString";
+import { wireObjectTypeV2ToSdkObjectDefinition } from "./wireObjectTypeV2ToSdkObjectDefinition";
+
+export async function generateClientSdkVersionTwoPointZero(
+  ontology: WireOntologyDefinition,
+  fs: MinimalFs,
+  outDir: string,
+) {
+  fs.writeFile(
```
pashakondratyev, line 28: Nit: Instead of minimalFS should we pass in a store to write files to act as an in memory file system and then we can parallelize the final outputting?

packages/generator/src/util/test/compileThis.ts
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+export function compileThis(files: Record<string, string>, basePath: string) {
+  const compilerOptions: ts.CompilerOptions = {
+    noEmit: true,
+    target: ts.ScriptTarget.ES2020,
```
mfedderly, line 22: we might want to make the generated typescript super strict so that anyone with lib type checking doesn't complain that we broke their build

packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.ts
ericanderson: Open to this. Can you do a followup PR?

packages/generator/src/util/test/compileThis.ts
ericanderson: Hmm. Do you have an example how this would manifest?

ericanderson:
I'm going to merge to unblock progress. Agree with feedback.

packages/generator/src/util/test/compileThis.ts
mfedderly: Reaching back into the cobweb-covered part of my brain I think its related to maybe when we turned on strict null checks? I think at some point there was a way for a library to be compiled with some insufficient level of strictness that could sometimes trip up consumers who wanted to enable the new strict modes before their dependencies got around to it. Maybe this just isn't a real concern in 2023.

https://github.com/microsoft/TypeScript/issues/14810#issuecomment-288811529

PR #24, opened 2023-10-24
Add legacy client stubs

ericanderson:
Effectively, this change takes a previously built internal sdk, copies and renames all the *.d.ts files to *.ts, fixes up import lines, resolves lint errors, and adds not implemented everywhere.

It is possible there are additional things in here that do not need to be ported but unfortunately submodule imports were possible on the existing v1.0. For now I just wanted to get the big working set up.

Still missing is the exported types from a generated `FoundryClient`. This will need to be stubbed out still by someone in a future PR.

There are also a number of optimizations that should be considered when working through these files. For example, there are a lot of exported `const foo = { a: () => {}, b: () => {} }` which can be treated instead as moved to a new file and the existing file doing `export * as foo from "./migratedFile"`. 

Run `./scripts/notImplemented.sh` from workspace root to see something like:
```
     1  packages/legacy-client/src/ontology-runtime/aggregations/ApproximateDistinctCountAggregatableProperty.ts:25:  throw new Error("not implemented");
...
   410  packages/legacy-client/src/ontology-runtime/aggregations/Aggregations.ts:352:  throw new Error("not implemented");
   411  packages/legacy-client/src/ontology-runtime/aggregations/Aggregations.ts:385:  throw new Error("not implemented");
```


ericanderson:
@pashakondratyev is going to take the first stab at the foundry client stubs in another pr.

pashakondratyev (approved):

PR #25, opened 2023-10-25
Adding stubs for FoundryClient for backcompat

pashakondratyev:
Adds stubs for `FoundryClient` and generating a back-compat interface. 
Here we're using a generated client to instantiate with the correct types from the metadata.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,5 +9,8 @@
   },
   "[json]": {
     "editor.defaultFormatter": "dprint.dprint"
+  },
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
   }
 }
--- a/packages/generator/src/v1.1/generateClientSdkVersionOneDotOne.test.ts
+++ b/packages/generator/src/v1.1/generateClientSdkVersionOneDotOne.test.ts
@@ -38,7 +38,10 @@ describe("generator", () => {
       helper.getFiles()[`${BASE_PATH}/index.ts`],
     ).toMatchInlineSnapshot(`
       "// Path: /foo/index.ts
+      import { Auth, FoundryClientOptions, BaseFoundryClient } from '@osdk/legacy-client';
       export const ontologyRid = 'ridHere';
+      export * from '@osdk/legacy-client';
+      export { FoundryClient } from './FoundryClient';
       "
     `);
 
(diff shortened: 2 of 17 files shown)


packages/generator/src/v1.1/generateClientSdkVersionOneDotOne.test.ts
```diff
@@ -49,6 +52,9 @@ describe("generator", () => {
       );
     }
 
-    expect(diagnostics).toHaveLength(0);
+    // TODO: Certain errors are expected since we can't resolve the static code, but we should fix them.
+    const errors = diagnostics.filter((q) => q.code !== 2792);
```
pashakondratyev, line 56: These are related to the imports in the generated code

.vscode/settings.json
```diff
@@ -9,5 +9,8 @@
   },
   "[json]": {
     "editor.defaultFormatter": "dprint.dprint"
+  },
+  "editor.codeActionsOnSave": {
```
pashakondratyev, line 13: Are there any strong opinions on adding this? I can add this to my user-settings instead

packages/legacy-client/src/metadata/OntologyDefinition.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { LocalDate } from "../ontology-runtime/baseTypes";
+
+export interface OntologyDefinition<K extends string> {
```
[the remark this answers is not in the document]
ericanderson: Yeah as we move forward we should consider how much we can consolidate.

.vscode/settings.json
ericanderson: I generally lean against forced on save actions. @mfedderly thoughts?

ericanderson (approved):

ericanderson:
Going to merge and we can handle follow ups separately if needed. Im happy for us to err on the side of moving quick.

.vscode/settings.json
mfedderly: I'm generally pro having robots do stuff for me. We have to pass eslint to get code merged anyhow so we should at least do this as a precommit hook. We can also turn this setting off super easily later if we don't like it.

PR #36, opened 2023-10-27
Adding basic interface generation and object set interfaces

pashakondratyev:


packages/generator/src/util/inMemoryFileSystem.ts
```diff
@@ -14,31 +14,19 @@
...
-  isPrimaryKey: boolean;
-  type: PropertyType;
-}
-
-interface PropertyType {
-  string: string;
-  datetime: LocalDate;
-  double: number;
-  boolean: boolean;
+export function getNewInMemoryFileSystem(): InMemoryFs {
```
ericanderson, line 23: nit: `createInMemoryFileSystem`

ericanderson (approved):
One nit otherwise loving where we are going.

PR #39, opened 2023-10-30
Adding select and OS Aggregations

pashakondratyev:
Adds Object Set Aggregations and Select syntax for loading specified properties from an ObjectSet.

### DIFF
--- a/packages/legacy-client/src/client/interfaces/aggregations.ts
+++ b/packages/legacy-client/src/client/interfaces/aggregations.ts
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2023 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


packages/legacy-client/src/client/interfaces/aggregations.ts
```diff
@@ -0,0 +1,127 @@
...
+) => Bucketing<K, V>;
+
+type IsGroupableProperty<T> = NonNullable<T> extends
+  number | LocalDate | Timestamp | string | boolean ? true : false;
+
+type GroupableProperties<T extends OntologyObject> = {
+  [K in keyof T as IsGroupableProperty<T[K]> extends true ? K : never]: T[K];
+};
+
+export declare type GroupBySelections<T extends OntologyObject> = {
```
mfedderly, line 67: I need to push my PR so you can see but this probably should be `<O extends OntologyDefinition<O>, K extends OntologyTypesFrom<O>>`. I've got a couple examples on my local for ObjectSet and friends

packages/legacy-client/src/client/interfaces/utils/OmitProperties.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { IsLink } from "./IsLink";
+
+export declare type OmitMetadataProperties<T> = {
```
ericanderson, line 19: We need to discuss shapes for this

ericanderson (approved):

PR #40, opened 2023-10-31
Add convertWireToOsdkObject to legacy-client

mfedderly:
`convertWireToOsdkObject` is the entry point for how we take data that came in over the wire and 'upgrade' it to have the rest of the methods expected to support the v1.1 apis. Its still *mostly* a stub at the moment.

In order to make this change, we actually had to do a bunch of work around the OntologyDefinition types.

The first change was to add `primaryKeyType` to the ObjectDefinition. Ideally we'd just define the name of the property and infer the typescript type of the primaryKey values, but that makes it really hard to infer in certain situations. We're open to revisiting this in a follow up.

The second change was to move some of the types which were using `OntologyObject` to `OntologyDefinition` plus an object type name to correctly infer types as you get further down a set of calls that traverse links.

### DIFF
--- a/packages/legacy-client/src/client/OsdkObject.ts
+++ b/packages/legacy-client/src/client/OsdkObject.ts
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2023 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type {
+  LinkDefinitionFrom,
+  LinkKeysFrom,
+  LinkTargetTypeFrom,
+  ObjectTypesFrom,
+  OntologyDefinition,
+  OsdkObjectPropertyType,
+  PropertyDefinitionFrom,
+  PropertyKeysFrom,
+} from "@osdk/api";
+import type {
+  MultiLink,
+  OntologyObject,
+  SingleLink,
+} from "../ontology-runtime/baseTypes";
+
+export type OsdkLegacyPropertiesFrom<
+  O extends OntologyDefinition<any>,
+  K extends ObjectTypesFrom<O>,
+> = {
+  [P in PropertyKeysFrom<O, K>]: OsdkObjectPropertyType<
+    PropertyDefinitionFrom<O, K, P>
+  >;
+};
+
+export type OsdkLegacyLinksFrom<
+  O extends OntologyDefinition<any>,
+  K extends ObjectTypesFrom<O>,
+> = {
+  [L in LinkKeysFrom<O, K>]: LinkTargetTypeFrom<O, K, L> extends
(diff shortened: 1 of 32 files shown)


packages/legacy-client/src/client/interfaces/filters.test.ts
```diff
@@ -20,15 +20,22 @@ import type {
...
+        complete: BooleanFilter;
+        body: StringFilter;
+        id: StringFilter;
+      }
+    >();
+    expectTypeOf<ObjectTypeFilter<typeof MockOntology, "Task">>().toMatchTypeOf<
+      {
+        id: NumericFilter;
+      }
+    >();
```
mfedderly, line 39: The compile errors for `.toMatchTypeOf` are really rough.

At first I thought that the old version was barfing because of a type error I made, but it was actually complaining because id should actually be a StringFilter given the current MockOntology. I added a second test case here to exercise the NumericFilter path.

packages/legacy-client/src/client/interfaces/objectSet.ts
```diff
@@ -14,54 +14,60 @@
...
-} from "../../ontology-runtime";
+import type { ObjectTypesFrom, OntologyDefinition } from "@osdk/api";
+import type { ListObjectsError, Page, Result } from "../../ontology-runtime";
+import type { OsdkLegacyObjectFrom } from "../OsdkObject";
 import type { ObjectTypeFilterFunction } from "./filters";
 import type { OrderByFunction } from "./ordering";
 import type { SearchAround } from "./searchAround";
 
-export type ObjectSet<O extends OntologyObject> =
+export type ObjectSet<
```
pashakondratyev, line 24: Should this type be internal? Something I'm concerned about is breaking the current experience of `const x : ObjectSet<Todo> = ...`  we may need to generate a bit more code to preserve some of the user facing types if we go this approach for these interfaces

packages/legacy-client/src/client/interfaces/filters.test.ts
pashakondratyev: Yeah it's a bit difficult to debug them, I've just ended up making objects like 
`x = {} as any as expectedType` and seeing what x looks like to see what's missing 😅

packages/legacy-client/src/client/interfaces/objectSet.ts
mfedderly: Yeah that's an interesting point, I guess we're kind of talking past each other a bit with the intention of these types. You intended them to be user-facing but I assumed they were going to be more geared to the output of the conversion process to make sure that all lined up.

Perhaps I should revert this set of changes and either 1) accept less type inference from the output step before it gets assigned the generated types, or 2) add a second set of types targeted towards the output before it gets wrapped.

@ericanderson how much do you care about the 'raw' output having type inference across links to different object types?

packages/legacy-client/src/client/OsdkObject.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+
+export type OsdkLegacyLinksFrom<
+  O extends OntologyDefinition<any>,
+  K extends ObjectTypesFrom<O>,
+> = {
+  [L in LinkKeysFrom<O, K>]: LinkTargetTypeFrom<O, K, L> extends
+    keyof O["objects"]
+    ? LinkDefinitionFrom<O, K, L>["multiplicity"] extends true
+      ? MultiLink<OsdkLegacyObjectFrom<O, LinkTargetTypeFrom<O, K, L>>>
```
mfedderly, line 49: Reverting the user-facing types gave us a small regression in the strictness of the primaryKey type for MultiLink. We can ignore for now and fix it up in a cleanup pass if we really want.

Type inference does seem to still work across link types at least, which is cool.

PR #41, opened 2023-10-31
SingleLinkImpl

mfedderly:
Lots of underneath the hood code churn to implement a relatively straightforward `SingleLinkImpl` class.

- Introduce `ClientContext` which is the object that gets passed around internally to preserve the information required to actually perform network requests.
- Add `ObjectsExceeded` to GetLinkedObjectError in order to make GetLinkedObjectsError assignable
- Start breaking out OntologyProvider into individual methods
- Rework of `getLinkedObject` logic in order to fix compile errors that were not found in the older implementation.

Skipping any refactoring of the error handling for now, we can do that in a later PR.

### DIFF
--- a/packages/legacy-client/src/client/convertWireToOsdkObject.ts
+++ b/packages/legacy-client/src/client/convertWireToOsdkObject.ts
@@ -24,23 +24,34 @@ import type {
   OntologyObject,
   ParameterValue,
 } from "../ontology-runtime/baseTypes";
+import { SingleLinkImpl } from "../ontology-runtime/baseTypes/SingleLinkImpl";
+import type { ClientContext } from "../ontology-runtime/ontologyProvider/calls/ClientContext";
 
 function createPrototype<
   T extends keyof O["objects"] & string,
   O extends OntologyDefinition<any>,
 >(
-  _primaryKey: ParameterValue,
+  context: ClientContext,
+  primaryKey: ParameterValue,
   type: T,
-  ontologyDefinition: O,
 ) {
-  const objDef = ontologyDefinition.objects[type];
+  const objDef = context.ontology.objects[type];
   const proto = {};
   for (
-    const [k, _v] of Object.entries(objDef.links)
+    const [k, { multiplicity, targetType }] of Object.entries(objDef.links)
   ) {
     Object.defineProperty(proto, k, {
       get: function() {
-        throw new Error("not implemented");
+        if (multiplicity == true) {
+          throw new Error("not implemented");
+        } else {
+          return new SingleLinkImpl(
+            context,
+            objDef.apiName,
+            primaryKey,
+            targetType,
+          );
+        }
       },
     });
   }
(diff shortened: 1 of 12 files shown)


packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/ClientContext.ts
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { OntologyDefinition, ThinClient } from "@osdk/api";
+import type { OntologyObjectV2 } from "@osdk/gateway/types";
+import type { OntologyObject } from "../..";
+
+export interface ClientContext {
```
mfedderly, line 21: This is intended to be the core of what you need to pass around in order to keep making network requests at any arbitrary depth of API calls.

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/getLinkedObject.ts
```diff
@@ -0,0 +1,60 @@
...
+    );
+  }
+
+  if (result.type == "ok" && result.value.length != 1) {
+    return createErrorResponse(
+      new Error(
+        `There are no ${targetObjectType} objects for this object`,
+      ),
+    );
+  }
```
mfedderly, line 57: The three error returns here don't pass typechecking, I'm not actually sure how you instantiate `GetLinkedObjectError` and none of those really match these error cases.

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/listLinkedObjects.ts
```diff
@@ -0,0 +1,67 @@
...
+import type { ClientContext } from "./ClientContext";
+import { getLinkedObjectsPage } from "./getLinkedObjectsPage";
+import { wrapResult } from "./util/wrapResult";
+
+export function listLinkedObjects<T extends OntologyObject>(
+  context: ClientContext,
+  sourceApiName: string,
+  primaryKey: any,
+  linkTypeApiName: T["__apiName"],
+): Promise<Result<T[], ListLinkedObjectsError>> {
```
mfedderly, line 31: Old style error handling still, but this breaks the listLinkedObjects function out of OntologyProvider.
Thoughts?

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -55,22 +66,22 @@ export function convertWireToOsdkObject<
   T extends keyof O["objects"] & string,
   O extends OntologyDefinition<any>,
 >(
+  context: ClientContext,
   obj: OsdkLegacyPropertiesFrom<O, T> & OntologyObject<T>,
-  ontologyDefinition: O,
 ): OsdkLegacyObjectFrom<O, T> {
-  const ontologyCache = cache.get(ontologyDefinition.metadata.ontologyRid);
+  const ontologyCache = cache.get(context.ontology.metadata.ontologyRid);
```
mfedderly, line 72: I'm considering that the first level of cache might be able to just be folded into `ClientContext`

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/listLinkedObjects.ts
pashakondratyev: Yeah this is much nicer than the big class!

pashakondratyev: We can refactor the error handling later

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/ClientContext.ts
```diff
@@ -0,0 +1,29 @@
...
+  client: ThinClient<any>;
+  ontology: OntologyDefinition<any>;
+  createObject<T extends OntologyObject>(
```
pashakondratyev, line 24: I think with the definition, we don't need createObject (in the existing world it was used to map back to the constructors from the generated code)

pashakondratyev (approved):

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/ClientContext.ts
mfedderly: Perhaps, I'm trying to figure out if there's benefits to passing this around or just relying on `convertWireToOsdkObject` as a global singleton.

pashakondratyev: We can go with this for now and always change it if we want

PR #42, opened 2023-10-31
Stub implementations for Objects

pashakondratyev:


packages/legacy-client/src/client/objectSets/OsdkObjectSet.ts
```diff
@@ -0,0 +1,99 @@
...
+export function createOsdkObjectSet<
+  O extends OntologyDefinition<any>,
+  K extends ObjectTypesFrom<O>,
+>(
+  objectSetDefinition: ObjectSetDefinition,
+  ontologyDefinition: O,
+): ObjectSet<OsdkLegacyObjectFrom<O, K>> {
+  const objectSet: ObjectSetOperations<OsdkLegacyObjectFrom<O, K>> = {
+    union(...otherObjectSets): ObjectSet<OsdkLegacyObjectFrom<O, K>> {
+      throw new Error("not implemented");
```
pashakondratyev, line 45: We may need the definitions to be a property of the ObjectSet so we can introspect into it

packages/legacy-client/src/client/ontology.ts
```diff
@@ -34,3 +43,20 @@ export class Ontology<O extends OntologyDefinition<any>> {
...
+    {
+      get: (_target, p, _receiver) => {
+        if (typeof p === "string") {
+          return createBaseOsdkObjectSet(clientContext, p, definition);
+        }
+
+        return undefined;
+      },
+    },
+  ) as Objects<O>;
```
mfedderly, line 61: As an alternate implementation you can skip the `Proxy` and just make an interface with a bunch of methods on it based on `definition`. Proxy is slower, but I'm also not sure that that super matters in the case where we're creating a whole object set vs hydrating a bunch of objects from the wire.

packages/legacy-client/src/client/objectSets/OsdkObjectSet.ts
```diff
@@ -0,0 +1,138 @@
...
+  ontologyDefinition: O,
+): BaseObjectSet<OsdkLegacyObjectFrom<O, K>> {
+  const baseObjectSetDefinition: BaseObjectSetDefinition = {
+    type: "base",
+    objectType: apiName,
+  };
+
+  const objectSet: BaseObjectSetOperations<OsdkLegacyObjectFrom<O, K>> = {
+    apiName: apiName as OsdkLegacyObjectFrom<O, K>["__apiName"],
+    description: "test",
```
mfedderly, line 121: Sneaky thing that should get fixed before release but isn't a `throw new Error()`.

mfedderly (approved):

packages/legacy-client/src/client/objectSets/OsdkObjectSet.ts
pashakondratyev: Added a new description property that we can use on the definition

PR #44, opened 2023-11-01
Implement object.toString for compatability

mfedderly:


packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -37,7 +37,24 @@ function createPrototype<
   type: T,
 ) {
   const objDef = context.ontology.objects[type];
-  const proto = { __apiName: type };
+  const proto = {};
+
+  Object.defineProperty(proto, "__apiName", { get: () => type });
```
mfedderly, line 42: This is now readonly

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -37,7 +37,24 @@ function createPrototype<
...
+
+  // toString that uses the ontology definition to enumerate the props that need to be serialized
+  proto.toString = function() {
+    const obj: Record<string, unknown> = {};
+    const self = this as OsdkLegacyPropertiesFrom<O, T> & OntologyObject<T>;
+    for (const prop of Object.keys(context.ontology.objects[type].properties)) {
+      obj[prop] = self[prop];
+      obj["__primaryKey"] = self.__primaryKey;
```
ericanderson, line 50: Copying these N times.

mfedderly: 🤦  fixed

PR #45, opened 2023-11-01
[WIP] Where clause and OrderBy

pashakondratyev:


packages/legacy-client/src/client/objectSets/OsdkObjectSet.ts
```diff
@@ -34,19 +35,22 @@ import { createObjectSetAggregationStep } from "./createObjectSetAggregationStep
...
+import { mapPropertiesToSearchFilter } from "./mapPropertiesToSearchFilter";
 
 export function createOsdkObjectSet<
   O extends OntologyDefinition<any>,
   K extends ObjectTypesFrom<O>,
 >(
   clientContext: ClientContext,
+  apiName: K,
   objectSetDefinition: ObjectSetDefinition,
   ontologyDefinition: O,
```
mfedderly, line 47: OntologyDefinition already lives in ClientContext, should we make it ClientContext<O> so that you can just use the fully type inferred ontology definition? (Future PR, I can do it)

mfedderly (approved):

packages/legacy-client/src/client/objectSets/OsdkObjectSet.ts
pashakondratyev: Yeah I actually realized we have it passed in three times when I was writing the test, the thin client, here, and context 🙈

PR #46, opened 2023-11-01
WIP support searchAround* methods on an ObjectSet

mfedderly:


packages/legacy-client/src/client/objectSets/OsdkObjectSet.ts
```diff
@@ -117,7 +126,7 @@ export function createBaseOsdkObjectSet<
   };
 
   const objectSet: BaseObjectSetOperations<OsdkLegacyObjectFrom<O, K>> = {
-    apiName: apiName as OsdkLegacyObjectFrom<O, K>["__apiName"],
+    apiName: apiName as string as OsdkLegacyObjectFrom<O, K>["__apiName"],
```
mfedderly, line 129: This cast was sad, but I think you can make this a `K`?

packages/legacy-client/src/client/objectSets/createObjectSetSearchAround.ts
```diff
@@ -15,12 +15,51 @@
...
+      } satisfies SearchAroundObjectSetDefinition;
+
+      const objSet = createOsdkObjectSet(
+        clientContext,
+        targetType,
+        definition,
+        ontologyDefinition,
+      );
+
+      return objSet as ObjectSet<O, any>;
```
mfedderly, line 56: For some reason this isn't satisfying the type on `result[key]` and I haven't figured it out yet

packages/legacy-client/src/client/interfaces/searchAround.ts
```diff
@@ -29,7 +29,7 @@ type InferLinkType<T> = T extends SingleLink<infer V> ? V
 export type SearchAround<T extends OntologyObject> = {
   [
     K in Extract<keyof T, string> as IsLink<T[K]> extends true
-      ? `searchAround${Capitalize<string & K>}`
+      ? `searchAround${Capitalize<K>}`
```
mfedderly, line 32: Extract<> above already enforces that K is a string

packages/legacy-client/src/client/objectSets/OsdkObjectSet.test.ts
```diff
@@ -53,4 +53,32 @@ describe("OsdkObjectSet", () => {
...
+      baseObjectSet,
+      MockOntology,
+    );
+
+    const searchAroundObjectSet = os.searchAroundTask();
+
+    expect(searchAroundObjectSet.definition).toEqual({
+      type: "searchAround",
+      objectSet: baseObjectSet,
+      link: "Task",
```
mfedderly, line 81: capital T in task is correct, right?

pashakondratyev: In this case yeah, It should be whatever the ApiName from the definition is

pashakondratyev (approved):

PR #47, opened 2023-11-02
Adding orderBy implementation, and terminal load steps

pashakondratyev:


packages/legacy-client/src/ontology-runtime/ontologyProvider/Errors.ts
```diff
@@ -58,12 +58,9 @@ export type GetObjectError =
...
   | PropertiesNotFound
   | CommonApiError;
-export type ListObjectsError =
-  | ObjectsExceededLimit
-  | ObjectTypeNotFound
-  | ObjectTypeNotSynced
-  | OntologySyncing
-  | CommonApiError;
+
+export type ListObjectsError = LoadObjectSetError;
```
pashakondratyev, line 62: This should be a non-breaking change to only have `LoadObjectSetError` since we do all those through one endpoint

packages/legacy-client/src/ontology-runtime/ontologyProvider/OntologyProvider.ts
```diff
@@ -354,7 +359,7 @@ export class OntologyProvider {
     propertyName: string,
     body: StreamPointsBody,
   ): AsyncGenerator<Result<TimeSeriesPoint<T>, TimeSeriesError>, any, unknown> {
-    yield* wrapIterator(
+    yield* this.wrapIterator(
```
mfedderly, line 362: I think you got a merge conflict, I already extracted wrapIterator, createPageIterator, and iterateLinkedObjects

packages/legacy-client/src/client/objectSets/createObjectSetOrderByStep.ts
```diff
@@ -15,12 +15,48 @@
...
+
+      return createObjectSetOrderByStep(
+        clientContext,
+        apiName,
+        objectSet,
+        ontologyDefinition,
+        [...orderByClauses, orderByClause],
+      );
+    },
+    ...createObjectSetTerminalLoadStep(
```
mfedderly, line 54: This is also added in OsdkObjectSet, did you mean to have it here also?

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/loadObjects.ts
```diff
@@ -0,0 +1,71 @@
...
+    objectSetDefinition,
+    orderByClauses,
+    selectedProperties,
+  );
+
+  if (page.type === "error") {
+    return page;
+  }
+
+  allObjects.push(...page.value.data);
```
mfedderly, line 50: Its potentially dangerous to do this if the length of .data is unbounded, you can go over the limit of the number of args in a function

```suggestion
  for (const object of page.value.data) {
    allObjects.push(object);
  }
```

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/loadObjects.ts
```diff
@@ -0,0 +1,71 @@
...
+      {
+        pageToken: page.value.nextPageToken,
+      },
+    );
+
+    if (page.type === "error") {
+      return page;
+    }
+
+    allObjects.push(...page.value.data);
```
mfedderly, line 67: Same here with the `.push(...page.value.data)`

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/loadObjects.ts
```diff
@@ -0,0 +1,71 @@
...
+    orderByClauses,
+    selectedProperties,
+  );
+
+  if (page.type === "error") {
+    return page;
+  }
+
+  allObjects.push(...page.value.data);
+  while (page.type === "ok" && page.value.nextPageToken) {
```
mfedderly, line 51: there's probably a cleaner way to do this with a `do{}while();` but it was like that when we got here.

pashakondratyev: Done 🫡

packages/legacy-client/src/client/objectSets/createObjectSetOrderByStep.ts
pashakondratyev: Yeah I needed a cleaner way of doing this, addressed!

mfedderly (approved):

PR #49, opened 2023-11-02
Aggregations

mfedderly:
Implement ObjectSet aggregations

- Implement createObjectSetAggregationStep and ComputeStep
- Extract the aggregate call from OntologyProvider
- Removes unused AggregatableObjectSet abstract class

Todo:
- [x] Compile errors 🤦 
- [x] Tests
- [x] Stricter types in mapPropertiesTo* reduce initial objects
- [ ] ~Cache the mapPropertiesTo* computation~ (will do in a separate PR)


packages/legacy-client/src/client/objectSets/mapPropertiesToMultipleAggregationProperties.ts
```diff
@@ -0,0 +1,56 @@
...
+            property,
+          );
+          break;
+
+        default:
+          const _: never = definition.type;
+      }
+      return acc;
+    },
+    { count: () => CountOperation } as Record<string, any>,
```
mfedderly, line 54: Did I read the generated code right that there's a sneaky `count` prop that gets added here?

pashakondratyev: Yeah that way we can also get a count of results in the aggregation responses

packages/legacy-client/src/client/objectSets/OsdkObjectSet.test.ts
```diff
@@ -75,6 +76,133 @@ describe("OsdkObjectSet", () => {
...
+    const os = createBaseTodoObjectSet(context);
+    mockAggregateResponse({ data: [] });
+    await os.count().compute();
+    expect(fetch).toHaveBeenCalledOnce();
+    expect(fetch).toHaveBeenCalledWith(
+      ...expectedJestResponse("Ontology/objectSets/aggregate", {
+        objectSet: { type: "base", objectType: "Todo" },
+        groupBy: [],
+        aggregation: [{ type: "count", name: "count" }],
+      }),
```
mfedderly, line 89: These tests are only testing that the request payload is correct, not that it handles the response. I can dig up the response payloads and update them later.

packages/legacy-client/src/client/objectSets/createObjectSetAggregationStep.ts
```diff
@@ -15,21 +15,173 @@
...
+        definition,
+        [
+          ...groupByClauses,
+          groupByClause,
+        ],
+      ) as AggregatableObjectSetStep<
+        AggregateSelection<OsdkLegacyObjectFrom<O, K>>,
+        MultipleAggregateSelection<OsdkLegacyObjectFrom<O, K>>,
+        GroupBySelections<OsdkLegacyObjectFrom<O, K>>,
+        any // TODO infer the TBucketGroup shape from groupByClause to be more strict here
```
mfedderly, line 130: This cast is kind of sad but I didn't want to get blocked on it

packages/legacy-client/src/client/objectSets/mapPropertiesToAggregatableProperties.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function mapPropertiesToAggregatableProperties<
+  O extends OntologyDefinition<any>,
+  K extends ObjectTypesFrom<O>,
+>(ontology: O, apiName: K) {
+  return Object.entries(
+    ontology.objects[apiName].properties,
+  ).reduce(
+    (acc, [property, _definition]) => {
+      // TODO use definitions to be more accurate here (exclude some properties and use a better metric for others)
```
mfedderly, line 31: This behavior matches the original implementation, we're protected by the generated interfaces right now.

packages/legacy-client/src/ontology-runtime/aggregations/ComputeStep.ts
```diff
@@ -14,14 +14,10 @@
...
-import {
-  type AggregateObjectsError,
-  type OntologyMetadata,
-  OntologyProvider,
-  type Result,
-} from "../ontologyProvider";
+import type { ObjectSetDefinition } from "../baseTypes";
+import { type AggregateObjectsError, type Result } from "../ontologyProvider";
+import { aggregate } from "../ontologyProvider/calls/aggregate";
+import type { ClientContext } from "../ontologyProvider/calls/ClientContext";
```
pashakondratyev, line 20: Not needed now but we should probably slowly reorganize the old code out of the `ontology-runtime` subfolder as we pull things in from it

pashakondratyev (approved):

PR #50, opened 2023-11-02
Adding support for more PropertyTypes

pashakondratyev:


examples/one_dot_one/src/generatedNoCheck/objects/ObjectTypeWithAllPropertyTypes.ts
```diff
@@ -0,0 +1,142 @@
...
+  GeoShape,
+  GeoPoint,
+  Attachment,
+  TimeSeries,
+} from '@osdk/legacy-client';
+
+/**
+ * A type with all property types
+ */
+export interface ObjectTypeWithAllPropertyTypes extends OntologyObject {
```
pashakondratyev, line 15: Just for illustrative purposes, can remove this one before we go forth with anything

packages/api/src/ontology/Definition.ts
```diff
@@ -88,14 +88,32 @@ export interface PropertyDefinition {
...
+  float: number;
+  decimal: number;
+  byte: number;
+
+  attachment: any;
+  geopoint: any;
+  geoshape: any;
 }
 
+export type AttachmentTypes = {
```
pashakondratyev, line 109: It felt like there are two ways to tackle this, one being have a `multiplicity` property on the property definition but then we'd need to infer the types on each one, or have string keys that map to the types. The reason I think multiplicity might not be the right solution here is we also will have a similar issue with TimeSeries which can have a string or a numeric subtype

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -58,6 +63,21 @@ function createPrototype<
...
+    //       const attachment = this.attachment;
+    //       Object.setPrototypeOf(
+    //         attachment,
+    //         AttachmentProperty(context, this.attachment),
+    //       );
+    //       return attachment;
+    //     },
+    //   });
+    // }
+  }
```
mfedderly, line 79: Did you mean to remove this block?

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -119,5 +138,133 @@ export function convertWireToOsdkObject<
   }
 
   Object.setPrototypeOf(obj, proto);
+  setPropertyPrototypes<T, O>(client, apiName, obj);
+
   return obj as OsdkLegacyObjectFrom<O, T>;
 }
+
+function setPropertyPrototypes<
```
mfedderly, line 146: `setPropertyAccessors`? Its not really touching the prototype

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -119,5 +138,133 @@ export function convertWireToOsdkObject<
...
+    }
+  }
+}
+
+function createPropertyPrototype<T>(
+  obj: OntologyObjectV2,
+  k: string,
+  constructor: (value: any) => T,
+) {
+  if (!obj[k]) {
```
mfedderly, line 213: This is a *little* sketchy because of things that can be falsy like `0` and `false`, but they don't get this far.

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -119,5 +138,133 @@ export function convertWireToOsdkObject<
...
+          if (!memoizedValue) {
+            memoizedValue = constructor(originalArray[index]);
+          }
+          return memoizedValue;
+        };
+      })(i),
+    });
+  }
+
+  Object.setPrototypeOf(slicedArray, Array.prototype);
```
mfedderly, line 246: Does it actually need this since it was created by `originalArray.slice()`?
I feel like there might be a simpler way to write this method, but I don't have a concrete suggestion at the moment.

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -119,5 +138,133 @@ export function convertWireToOsdkObject<
...
+  key: string,
+  constructor: (value: any) => T,
+) {
+  const originalValue = object[key];
+  Object.defineProperty(object, key, {
+    get: (function(): () => T {
+      let memoizedValue: T | undefined;
+      return function() {
+        if (!memoizedValue) {
+          console.log(originalValue);
```
mfedderly, line 261: We should ban console.log at the eslint level so it doesn't sneak in

packages/legacy-client/src/client/objectSets/mapPropertiesToMultipleAggregationProperties.ts
```diff
@@ -35,11 +35,22 @@ export function mapPropertiesToMultipleAggregationProperties<
...
+        case "long":
+        case "short":
           acc[property] = NumericPropertyMetric(property);
 
+        case "byte":
+        case "attachment":
+        case "geopoint":
+        case "geoshape":
         case "boolean":
         case "string":
```
mfedderly, line 55: I was kind of surprised to see byte and boolean down here instead of the block with the other numeric types.

packages/legacy-client/src/ontology-runtime/baseTypes/attachments/AttachmentProperty.ts
```diff
@@ -14,81 +14,28 @@
...
-      rid: string;
+export const AttachmentProperty = (
+  thinClient: ThinClient<any>,
+  attachmentRid: string,
+): Attachment => {
+  return {
+    attachmentRid: attachmentRid,
+    type: "Attachment" as const,
+    getMetadata() {
+      console.log("AttachmentProperty.getMetadata");
```
mfedderly, line 34: another console.log

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/getAttachment.ts
```diff
@@ -0,0 +1,42 @@
...
+import {
+  AttachmentsErrorHandler,
+  handleAttachmentsError,
+} from "../ErrorHandlers";
+import type { AttachmentsError } from "../Errors";
+import type { Result } from "../Result";
+import { wrapResult } from "./util/wrapResult";
+
+export async function getAttachment(
+  client: ThinClient<any>,
```
mfedderly, line 28: I need to make a pass but this actually should be `ThinClient<OntologyDefinition<any>>`

packages/legacy-client/src/client/convertWireToOsdkObject.ts
pashakondratyev: Yeah this one is not needed actually

packages/legacy-client/src/client/objectSets/mapPropertiesToMultipleAggregationProperties.ts
pashakondratyev: Byte is a good catch, though I'm not sure if it is aggregatable

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -119,5 +123,131 @@ export function convertWireToOsdkObject<
...
 }
+
+function setPropertyAccessors<
+  T extends ObjectTypesFrom<O> & string,
+  O extends OntologyDefinition<any>,
+>(client: ThinClient<O>, apiName: T, obj: OntologyObjectV2) {
+  for (
+    const [k, v] of Object.entries(client.ontology.objects[apiName].properties)
+  ) {
+    if (!obj[k]) {
```
mfedderly, line 138: You probably want to explicitly check against null here because of values like `0`

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -119,5 +123,131 @@ export function convertWireToOsdkObject<
...
+      let memoizedValue: T | undefined;
+      return function() {
+        if (!memoizedValue) {
+          memoizedValue = constructor(originalValue);
+        }
+        return memoizedValue;
+      };
+    })(),
+    set(v) {
+    },
```
mfedderly, line 251: Do we need a set implementation?

mfedderly (approved):

PR #52, opened 2023-11-02
Remove ClientContext

mfedderly:
This was an unnecessary wrapper around calls to `convertWireToOsdkObject` and `ThinClient`


packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -87,29 +91,30 @@ function createPrototype<
...
 export function convertWireToOsdkObject<
-  T extends keyof O["objects"] & string,
+  T extends ObjectTypesFrom<O> & string,
   O extends OntologyDefinition<any>,
 >(
-  context: ClientContext,
-  obj: OsdkLegacyPropertiesFrom<O, T> & OntologyObject<T>,
+  client: ThinClient<O>,
+  apiName: T,
+  obj: OntologyObjectV2,
```
mfedderly, line 99: This change right here is pretty important. It effectively gives us `Record<string, any>`, and we're depending on it to contain `__rid` and `__primaryKey` on it.

packages/legacy-client/src/client/convertWireToOsdkObject.ts
```diff
@@ -87,29 +91,30 @@ function createPrototype<
...
 ): OsdkLegacyObjectFrom<O, T> {
-  const ontologyCache = cache.get(context.ontology.metadata.ontologyRid);
-  let proto = ontologyCache?.get(obj.__apiName);
+  const ontologyCache = cache.get(client.ontology.metadata.ontologyRid);
+  let proto = ontologyCache?.get(apiName);
 
   if (proto == null) {
-    let proto = createPrototype(
-      context,
+    proto = createPrototype(
```
mfedderly, line 105: 🤦 shadowed my proto variable

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/getObject.ts
```diff
@@ -14,32 +14,33 @@
...
 import type { OntologyObject } from "../../baseTypes";
 import { GetObjectErrorHandler, handleGetObjectError } from "../ErrorHandlers";
 import type { GetObjectError } from "../Errors";
 import type { Result } from "../Result";
-import type { ClientContext } from "./ClientContext";
 import { wrapResult } from "./util/wrapResult";
 
 export async function getObject<T extends OntologyObject>(
-  context: ClientContext,
+  client: ThinClient<any>,
```
pashakondratyev, line 28: I can take care of this in a follow up but we can probably change all the calls to take the OntologyDefinition as a type and use that here

packages/legacy-client/src/client/convertWireToOsdkObject.ts
pashakondratyev: Yeah unfortunately we need to make that assumption at some point

pashakondratyev (approved):

PR #53, opened 2023-11-03
ThinClient cleanup

mfedderly:
- `ThinClient<any>` should really be `ThinClient<OntologyDefinition<any>>`
- Delete `get` from OntologyProvider (it is `getObjects()` now)
- Which descended into fixing up auth/fetchFn/stack accesses in a bunch of places
- The `ThinClient<any>` change required some additional work to make the types all happy

### DIFF
--- a/packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/util/iterateLinkedObjects.ts
+++ b/packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/util/iterateLinkedObjects.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import type { ThinClient } from "@osdk/api";
+import type { OntologyDefinition, ThinClient } from "@osdk/api";
 import { createOpenApiRequest } from "@osdk/api";
 import { listLinkedObjectsV2 } from "@osdk/gateway/requests";
 import type { ListLinkedObjectsResponseV2 } from "@osdk/gateway/types";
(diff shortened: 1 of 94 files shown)


packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/util/iterateLinkedObjects.ts
```diff
@@ -43,8 +43,11 @@ export async function* iterateLinkedObjects<T extends OntologyObject>(
       },
     );
 
+    const { targetType } =
+      client.ontology.objects[sourceApiName].links[linkTypeApiName];
```
mfedderly, line 47: Actual bugfix - use the correct targetType instead of the link type when we convert the linked objects

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/util/iterateLinkedObjects.ts
```diff
@@ -43,8 +43,11 @@ export async function* iterateLinkedObjects<T extends OntologyObject>(
...
+
     for (const object of page.data) {
-      yield convertWireToOsdkObject(client, linkTypeApiName, object);
+      yield convertWireToOsdkObject(client, targetType, object) as unknown as T;
```
mfedderly, line 50: I'm a little sad about the `as unknown` cast here

pashakondratyev: Would writing it as: 
`convertWireToOsdkObject<T["__apiName">, typeof client.OntologyDefinition>` or something along those lines be better?

mfedderly: That didn't work either, the generics for T are unassignable, probably because this is a seam between `<T extends OntologyObject>` and the OntologyDefinition side.

pashakondratyev: Ah I see, I think this is fine for now then and we can come back to it?

pashakondratyev (approved):

PR #54, opened 2023-11-03
Implement ObjectSet.select

mfedderly:


packages/legacy-client/src/client/objectSets/OsdkObjectSet.ts
```diff
@@ -108,8 +109,16 @@ export function createOsdkObjectSet<
...
+    ): FilteredPropertiesTerminalOperationsWithGet<
+      OsdkLegacyObjectFrom<O, K>,
+      T[]
+    > {
+      return createFilteredPropertiesObjectSetWithGetTerminalOperationsStep(
+        client,
+        apiName,
+        objectSetDefinition,
+        properties,
+      );
```
mfedderly, line 121: This returned object has a `get` method on it, which is guarded against by the interfaces. I can rework the component parts to remove it if you think that's important.

packages/legacy-client/src/client/interfaces/baseObjectSet.ts
```diff
@@ -26,8 +26,8 @@ import type { ObjectSet } from "./objectSet";
 import type { SelectableProperties } from "./utils/OmitProperties";
 
 export type BaseObjectSet<O extends OntologyObject> =
-  & ObjectSet<O>
-  & BaseObjectSetOperations<O>;
+  & BaseObjectSetOperations<O>
+  & ObjectSet<O>;
```
mfedderly, line 30: Changing the order here seemed to give precedence to the first `select` typings it found, which gives us the `...WithGet` return type there.

packages/legacy-client/src/client/objectSets/OsdkObjectSet.test.ts
```diff
@@ -262,10 +320,9 @@ const baseObjectSet: ObjectSetDefinition = {
 };
 
 function createBaseTodoObjectSet(client: ThinClient<any>) {
-  const os = createOsdkObjectSet<typeof MockOntology, "Todo">(
+  const os = createBaseOsdkObjectSet<typeof MockOntology, "Todo">(
```
mfedderly, line 323: Gave us the correct type branding for a base object set for the select

pashakondratyev (approved):

pashakondratyev (approved):

PR #55, opened 2023-11-03
Use eslint to guard against shipping console.log statements accidentally

mfedderly:
Do you agree with the default rule and how I added the exceptions in the few cases that needed it?

### DIFF
--- a/monorepo/eslint-config-sane/library.cjs
+++ b/monorepo/eslint-config-sane/library.cjs
@@ -63,6 +63,7 @@ module.exports = {
     "import/no-unresolved": "off",
     "import/no-named-as-default": "error",
     "unused-imports/no-unused-imports": "error",
+    "no-console": "error",
   },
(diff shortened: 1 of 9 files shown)


monorepo/eslint-config-sane/library.cjs
```diff
@@ -73,5 +74,11 @@ module.exports = {
     {
       files: ["*.ts", "*.tsx", "*.mts", "*.cts"],
     },
+    {
+      files: ["**/*.test.ts", "**/test/*"],
```
mfedderly, line 78: hopefully nobody ever names a directory `test` and ships code out of it 😆

ericanderson (approved):

monorepo/eslint-config-sane/library.cjs
mfedderly: Yeah we're using it `packages/generator/src/util/test/createMockMinimalFiles.ts`

PR #61, opened 2023-11-06
Adding Action interfaces and metadata

pashakondratyev:


packages/api/src/ontology/Definition.ts
```diff
@@ -40,11 +40,45 @@ export type PropertyDefinitionFrom<
...
+  integer: number;
+  timestamp: Date;
+  short: number;
+  long: number;
+  float: number;
+  decimal: number;
+  byte: number;
+}
+
+export interface ActionParameterDefinition<K> {
```
mfedderly, line 77: Are we get into trouble where `K` cannot be known to be in `ObjectTypesFrom<O>`?

packages/legacy-client/src/client/actions.ts
```diff
@@ -0,0 +1,93 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ObjectTypesFrom, OntologyDefinition } from "@osdk/api";
+import type { ObjectSet } from "./interfaces";
+import type { OsdkLegacyObjectFrom } from "./OsdkObject";
+
+export interface ValidActionParameterTypes {
```
mfedderly, line 21: This is defined twice now

pashakondratyev: Yeah this one needs to map to the legacy types, need to fix that

packages/api/src/ontology/Definition.ts
pashakondratyev: I think it is a valid assumption that and ObjectType used by an action must exist in the definition

mfedderly: Yeah I'm less worried about the data itself and more proving that to the type system

pashakondratyev: Will improve that!

examples/one_dot_one/src/generatedNoCheck/objects/ObjectTypeWithAllPropertyTypes.ts
```diff
@@ -56,118 +56,147 @@ export const ObjectTypeWithAllPropertyTypes = {
     id: {
       multiplicity: false,
       type: 'integer',
+      nullable: true,
```
mfedderly, line 59: Did you mean to make all of these nullable?

packages/legacy-client/src/client/OsdkObject.ts
```diff
@@ -92,5 +92,8 @@ export type OsdkLegacyObjectFrom<
 > = K extends string ?
     & OsdkLegacyPropertiesFrom<O, K>
     & OsdkLegacyLinksFrom<O, K>
-    & OntologyObject<K>
+    & OntologyObject<
+      K,
+      ValidLegacyPropertyTypes[O["objects"][K]["primaryKeyType"]]
```
mfedderly, line 97: We should probably add a `PrimaryKeyTypeOf<O,K>` helper

packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface ActionDefinition<
+  A extends String,
+  K extends String,
```
mfedderly, line 19: probably lowercase `string` here?

packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -0,0 +1,46 @@
...
+> {
+  apiName: A;
+  description?: string;
+  displayName?: string;
+  parameters: Record<string, ActionParameterDefinition<K>>;
```
mfedderly, line 24: Can we have different K's on different parameters?

packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface ActionDefinition<
+  A extends String,
+  K extends String,
+> {
+  apiName: A;
```
mfedderly, line 21: At least for queries, these were (almost) branded strings. Is there a reason why we didn't keep that going forward? I already caught some bugs I made when I tried to use a LinkApiName instead of an ObjectApiName, for instance.

packages/api/src/ontology/ActionDefinition.ts
pashakondratyev: Yup!

examples/one_dot_one/src/generatedNoCheck/objects/ObjectTypeWithAllPropertyTypes.ts
pashakondratyev: Yeah, we don't have this metadata at the moment so they should all be nullable

packages/generator/src/v1.1/generateMetadataFile.test.ts
```diff
@@ -38,8 +38,10 @@ describe(generateMetadataFile, () => {
       "import type { OntologyDefinition } from '@osdk/api';
       import type { Ontology as ClientOntology } from '@osdk/legacy-client';
       import type { Objects } from './ontologyObjects';
+      import type { Actions } from './ontologyActions';
       import { Todo } from './objects/Todo';
       import { Person } from './objects/Person';
+      import { markTodoCompleted } from './actions/markTodoCompleted';
```
pashakondratyev, line 44: Something to watch out for and track is to make sure we don't conflict with the imports

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/getLinkedObject.ts
```diff
@@ -47,7 +47,11 @@ export function getLinkedObject<T extends OntologyObject>(
           select: [],
         },
       );
-      return convertWireToOsdkObject(client, linkTypeApiName, object) as T;
+      return convertWireToOsdkObject(
+        client,
+        linkTypeApiName,
+        object,
+      ) as unknown as T;
```
pashakondratyev, line 54: This is a side effect of having `OntologyObject` be parametrized with parameter value, I'm tracking cleaning these up in a separate branch

packages/api/src/ontology/ActionDefinition.ts
mfedderly: Oh right but K can be several strings not just one. I always forget that.

packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export interface ValidBaseActionParameterTypes {
+  boolean: boolean;
+  string: string;
+  integer: number;
+  long: number;
+  double: number;
+  datetime: Date;
+  timestamp: Date;
+  attachment: any;
```
mfedderly, line 41: `Attachment` instead of any?

packages/generator/src/shared/wireActionTypeV2ToSdkActionDefinition.ts
```diff
@@ -0,0 +1,124 @@
...
+  input: ActionTypeV2,
+): ActionDefinition<any, any> {
+  const modifiedEntityTypes = getModifiedEntityTypes(input);
+  return {
+    apiName: input.apiName,
+    parameters: Object.fromEntries(
+      Object.entries(input.parameters).map((
+        [key, value],
+      ) => [key, wireActionParameterV2ToSdkParameterDefinition(value)]),
+    ),
```
mfedderly, line 39: I really like `Object.fromEntries` for this

packages/generator/src/shared/wireActionTypeV2ToSdkActionDefinition.ts
```diff
@@ -0,0 +1,124 @@
...
+    ActionModifiedEntity
+  >;
+
+  for (const key of addedObjects) {
+    entities[key] = { created: true, modified: false };
+  }
+
+  for (const key of modifiedObjects) {
+    if (entities[key]) {
+      entities[key] = { ...entities[key], modified: true };
```
mfedderly, line 117: you could just `entities[key].modified = true;`

packages/generator/src/v1.1/generateActions.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    let parameterBlock = "";
+    if (entries.length > 0) {
+      parameterBlock = `params: { \n`;
+      for (
+        const [parameterName, parameterData] of entries
+      ) {
+        parameterBlock += `"${parameterName}"`;
+        parameterBlock += parameterData.required ? ": " : "?: ";
+        const getTypeScriptType = handleDataType(
```
mfedderly, line 52: `getTypeScriptName` is a little weird naming wise

packages/generator/src/v1.1/generateActions.ts
```diff
@@ -0,0 +1,133 @@
...
+      importedObjects.add(objectType);
+      return `ObjectSet<${objectType}>`;
+    }
+    case "object": {
+      const objectType = actionParameter.objectTypeApiName!;
+      importedObjects.add(objectType);
+      return `${objectType}`;
+    }
+    case "array":
+      return handleDataType(actionParameter.subType, importedObjects) + `[]`;
```
mfedderly, line 114: I did this as `Array<${innerType}>` just in case I got a union, the `[]` would be sort of ambiguous. I don't think you have that problem here though.

mfedderly (approved):

packages/generator/src/v1.1/generateActions.ts
pashakondratyev: 🙈

packages/api/src/ontology/ActionDefinition.ts
pashakondratyev: From offline: Saving the definitions here and in a few other places until we've decided on a 2.0 type

PR #65, opened 2023-11-07
Add deployment to site

ericanderson:
image


packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -0,0 +1,65 @@
...
+  builder: (argv) => {
+    return argv
+      .options({
+        "siteVersion": {
+          type: "string",
+          conflicts: "clearVersion",
+          // group: "Deploy Version",
+          // implies: { "clearVersion": "false" },
+        },
+        clearVersion: {
```
mfedderly, line 36: The meaning of `clearVersion` isn't clear to me. I'm assuming it just undeploys your app entirely or something? its worth docs or a rename or something even though the underlying APIs say clearVersion.

mfedderly (commented):
the fancy console coloring is neat

mfedderly (approved):

PR #69, opened 2023-11-09
Query legacy-client impl WIP

mfedderly:


packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/executeQuery.ts
```diff
@@ -0,0 +1,415 @@
...
+    return value.objectSetDefinition;
+  } else if (typeof value === "object") {
+    // Since structs are valid arguments for Queries, we map the values
+    return Object.entries(value).reduce((acc, [key, structValue]) => {
+      acc[key] = getParameterValueMapping(structValue);
+      return acc;
+    }, {} as { [key: string]: PrimitiveParameterValue });
+  }
+
+  return value as string | number | boolean;
```
mfedderly, line 139: This return type is a bit of a lie, you can get Array<any>, Record<string, any>, and an ObjectSetDefinition

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/executeQuery.ts
```diff
@@ -0,0 +1,415 @@
...
+): obj is { objectSetDefinition: ObjectSetDefinition } {
+  return obj && obj.objectSetDefinition;
+}
+
+async function remapQueryResponseType(
+  client: ThinClient<OntologyDefinition<any>>,
+  definition: QueryDataTypeDefinition<any>,
+  responseValue: PrimitiveParameterValue,
+): Promise<ParameterValue> {
+  // TODO can the backend really not send us null responses?
```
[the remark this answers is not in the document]
pashakondratyev: Yeah that would be good

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/executeQuery.ts
```diff
@@ -0,0 +1,415 @@
...
+            };
+          });
+
+          return {
+            groups,
+          };
+        }
+
+        case "objectSet":
+          throw new Error("ObjectSet type not supported in response");
```
pashakondratyev, line 312: We can handle this separately in a FLUP, but we can probably fill this in, just never got around to this in V0

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/executeQuery.ts
```diff
@@ -0,0 +1,415 @@
...
+      acc[key] = getParameterValueMapping(value);
+      return acc;
+    },
+    parameterMap,
+  );
+
+  return remappedParams;
+}
+
+function getParameterValueMapping(
```
pashakondratyev, line 110: I think we can share this code with the code in `executeAction`

packages/legacy-client/src/client/queries.test.ts
```diff
@@ -0,0 +1,131 @@
...
+          "integer": 0,
+          "long": 0,
+          "boolean": false,
+          "date": nowLocalDate.toISOString(),
+          "string": "",
+          "timestamp": nowTimestamp.toISOString(),
+          "object": {},
+          "objectSet": {
+            "apiName": "Todo",
+            "description": "",
```
pashakondratyev, line 106: I don't think description should be here

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/executeQuery.ts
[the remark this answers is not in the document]
mfedderly: Lets move it into a flup PR, when I start to mess with the types things seem to be breaking in unexpected ways.

pashakondratyev (approved):

PR #71, opened 2023-11-09
Adding timeseries support and fixing some nits

pashakondratyev:


packages/generator/src/shared/wirePropertyV2ToSdkPrimaryKeyTypeDefinition.ts
```diff
@@ -23,13 +23,33 @@ export function wirePropertyV2ToSdkPrimaryKeyTypeDefinition(
...
+    case "float":
+    case "geopoint":
+    case "geoshape":
+    case "long":
+    case "short": {
       return input.dataType.type;
     }
-    default: {
+    case "date": {
+      return "datetime";
```
mfedderly, line 39:
```suggestion
      return "localdate";
```
may require a bunch of downstream changes though

packages/legacy-client/src/ontology-runtime/ontologyProvider/calls/iterateTimeSeriesPoints.ts
```diff
@@ -0,0 +1,87 @@
...
+    e => handleTimeSeriesError(new TimeSeriesErrorHandler(), e, e.parameters),
+  );
+}
+
+async function processStream<T extends string | number>(
+  streamIterator: AsyncGenerator<any, any, unknown>,
+): Promise<any[]> {
+  const allPoints: Array<TimeSeriesPoint<T>> = [];
+  for await (const points of streamIterator) {
+    Array.prototype.push.apply(allPoints, points);
```
mfedderly, line 84:
```suggestion
    for (const point of points) {
        allPoints.push(point);
    }
```

to avoid the unbounded number of args

mfedderly (approved):

packages/generator/src/shared/wirePropertyV2ToSdkPrimaryKeyTypeDefinition.ts
pashakondratyev: Lets revisit this one, we can also rename it to be consistent with the type in the ontology metadata (just `date`)

PR #73, opened 2023-11-13
Fixing bugs in integration

pashakondratyev:


packages/legacy-client/src/client/objectSets/OsdkObjectSet.test.ts
```diff
@@ -243,12 +245,12 @@ describe("OsdkObjectSet", () => {
         pageToken: "fakePageToken",
       }),
     );
-    expect(result.type).toEqual("ok");
+    const selectedObjects = unwrapResultOrThrow(result);
```
mfedderly, line 248: Its kind of sad that `selectedObjects` is unused and also doesn't have a `_` prefix

packages/legacy-client/src/client/objects/convertWireToOsdkObject.ts
```diff
@@ -68,7 +69,7 @@ function createPrototype<
 
   // add the relevant keys for the link types associated with this object type
   for (
-    const [k, { multiplicity, targetType }] of Object.entries(objDef.links)
+    const [k, { multiplicity }] of Object.entries(objDef.links)
```
mfedderly, line 72: Can we stop serializing `targetType` entirely into the OntologyDefinition? I guess we still use it for type inference?

packages/generator/src/util/test/createMockMinimalFiles.ts
```diff
@@ -32,7 +32,6 @@ export function createMockMinimalFiles() {
     dumpFilesToConsole: () => {
       for (const [path, contents] of Object.entries(getFiles())) {
         console.group(path);
-        console.log(contents);
```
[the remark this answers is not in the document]
pashakondratyev: Oh yeah, we can probably put that back in now

packages/legacy-client/src/client/objects/convertWireToOsdkObject.ts
pashakondratyev: Yeah I think we'd still need it for type inference, and also just readability of the definition

packages/legacy-client/src/client/objectSets/OsdkObjectSet.test.ts
pashakondratyev: Was using that to be able to see what the error was quickly, but probably should have a eslint rule for no-unused-vars

mfedderly (approved):

PR #74, opened 2023-11-13
Queries follow-ups

mfedderly:
- Add ObjectSet and nullable implementations to query response transforms
- Remove `as any` cast from executeQuery


pashakondratyev:
Some bugs I caught in the testing:
https://github.com/pashakondratyev/osdk-ts/blob/715c1dc61447817cd9ba00deb6c932fe84b7a9ac/packages/legacy-client/src/client/net/executeQuery.ts#L297
and 
` Cannot remapQueryBucketKeyType with unsupported type {"keyType":"string","valueType":"double"}
`Cannot remapQueryBucketKeyType with unsupported type {"keyType":"string","valueType":{"keyType":"range","keySubtype":"timestamp","valueType":"double"}}`

pashakondratyev (approved):

PR #75, opened 2023-11-13
Rework convertWireToOsdkObject caching

mfedderly:
Caching the object prototypes with an implied client means that multiple clients for the same ontology could accidentally use the incorrect client.

Instead we use the same prototype but pass through the ThinClient reference as a non-enumerable symbol on the object itself.


pashakondratyev (approved):

ericanderson:
Nice

PR #82, opened 2023-11-14
Addressing bugs in generated code

pashakondratyev:


examples/one_dot_one/src/generatedNoCheck/ontologyQueries.ts
```diff
@@ -50,7 +50,7 @@ export interface Queries {
     array: Array<string>;
     set: Set<string>;
     unionNonNullable: string | number;
-    unionNullable?: string | number | null;
+    unionNullable?: string | number;
```
mfedderly, line 53: Is there a missing change in this PR that makes these changes?

packages/generator/src/v1.1/generateQueries.ts
```diff
@@ -147,9 +147,9 @@ function handleQueryDataType(
       return `{ ${properties.join(",\n")} }`;
 
     case "union":
-      return dataType.unionTypes.map(type =>
-        handleQueryDataType(type, importedObjects)
-      ).join("|");
+      return dataType.unionTypes.filter(subType =>
+        !isNullableQueryDataType(subType)
+      ).map(type => handleQueryDataType(type, importedObjects)).join("|");
```
mfedderly, line 152: It might be safer to do the filter step last instead of first and just delete "null".

union(string, union(number, null)) <- something like this

mfedderly (approved):

PR #83, opened 2023-11-14
Add 2.0 dates and geo

ericanderson:
Based on feedback, getting dates converted and geo

### DIFF
--- a/packages/api/src/client/object/convertWireToOsdkObjects.ts
+++ b/packages/api/src/client/object/convertWireToOsdkObjects.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import { Temporal } from "@js-temporal/polyfill";
 import type { OntologyObjectV2 } from "@osdk/gateway/types";
 import type {
   ObjectTypesFrom,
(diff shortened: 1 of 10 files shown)


packages/api/src/client/object/convertWireToOsdkObjects.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { Temporal } from "@js-temporal/polyfill";
```
mfedderly, line 17: Using `Temporal` kind of sketches me out. I haven't checked the size of the polyfill (which depends on the bigint polyfill too), but its also tying us to the spec which is still in the proposal stage which reminds me of the time that typescript added decorators and got bit. No browsers claim to support this yet.

ericanderson: TS added decorators at stage 1. This is stage 3 where it has been for awhile and the last thing they wanted to stabilize just closed.

packages/api/src/ontology/Definition.ts
```diff
@@ -118,26 +122,31 @@ export interface ValidPropertyTypes {
   stringTimeseries: any;
 
   attachment: any;
-  geopoint: any;
-  geoshape: any;
+  geopoint: GeoJSON.Point;
+  geoshape: GeoJSON.Geometry;
```
mfedderly, line 126: Do you want to allow GeometryCollections? I'm also really interested at comparing BSON size/decode speed for geojson stuff across the wire.

packages/api/src/ontology/Definition.ts
```diff
@@ -118,26 +122,31 @@ export interface ValidPropertyTypes {
...
-  true ? Array<U> : U;
+type MaybeArray<T extends { multiplicity?: boolean | undefined }, U> =
+  T["multiplicity"] extends true ? Array<U> : U;
 
 type MaybeNullable<T extends PropertyDefinition, U> = T["nullable"] extends true
   ? U | undefined
   : U;
 
+type Raw<T> = T extends Array<any> ? T[0] : T;
+type Converted<T> = T extends Array<any> ? T[1] : T;
```
mfedderly, line 137: These feel weird just picking arbitrary indicies out of an array, but I'm assuming they're used to pull things out of a configuration metadata?

packages/api/src/ontology/Definition.ts
pashakondratyev: I believe GeometryCollections are supported but Feature and Feature collections are not

mfedderly (approved):

packages/api/src/ontology/Definition.ts
ericanderson: This was from when i was trying to do something clever and left it in for now.

PR #85, opened 2023-11-15
Adding reserved keywords logic for the internal legacy types

pashakondratyev:


packages/legacy-client/src/client/interfaces/aggregations.ts
```diff
@@ -107,7 +107,7 @@ export declare type MultipleAggregateSelection<T extends OntologyObject> =
     count: () => CountOperation;
   };
 
-type FilterFromType<T, N extends string> = NonNullable<T> extends number
+type GroupByFromType<T, N extends string> = NonNullable<T> extends number
```
pashakondratyev, line 110: The `N` here is still problematic if you do a groupBy the type you get is `groups: { group: { `count_`: ... } }[]` but I think we can tackle that one separately. I don't think the existing would work at all in this case

mfedderly (approved):

PR #90, opened 2023-11-15
Generated conjure-lite functions

ericanderson:
I didn't want to keep hand writing conjure functions. This step is just to replace the ones I wrote by hand. Then we can start to use this for other things in our roadmap

I broke these up into rebased commits to make it easier to review.


packages/cli/package.json
```diff
@@ -46,6 +46,9 @@
   "publishConfig": {
     "access": "public"
   },
+  "imports": {
+    "#net": "./src/net/index.mts"
```
mfedderly, line 50: Kind of wild that you can import from a .mts file

mfedderly (approved):

packages/cli/package.json
ericanderson: its a bit broken honestly. @pashakondratyev has tickets. we have to be careful with this mechanism.

PR #92, opened 2023-11-16
Making package structure to match legacy code

pashakondratyev:


examples/one_dot_one/src/generatedNoCheck/Ontology.ts
```diff
@@ -1,15 +1,15 @@
...
-import { Person } from './objects/Person';
-import { Todo } from './objects/Todo';
+import { actionTakesAllParameterTypes } from './ontology/actions/actionTakesAllParameterTypes';
+import { createTodo } from './ontology/actions/createTodo';
+import { ObjectTypeWithAllPropertyTypes } from './ontology/objects/ObjectTypeWithAllPropertyTypes';
+import { Person } from './ontology/objects/Person';
+import { Todo } from './ontology/objects/Todo';
+import { getTodoCount } from './ontology/queries/getTodoCount';
+import { queryTakesAllParameterTypes } from './ontology/queries/queryTakesAllParameterTypes';
 import type { Actions } from './ontologyActions';
```
pashakondratyev, line 10: I wonder if we should move these into `ontology/actions` as well

examples/one_dot_one/src/generatedNoCheck/Ontology.ts
```diff
@@ -1,15 +1,15 @@
...
 import type { Objects } from './ontologyObjects';
 import type { Queries } from './ontologyQueries';
-import { getTodoCount } from './queries/getTodoCount';
-import { queryTakesAllParameterTypes } from './queries/queryTakesAllParameterTypes';
 
 export const Ontology: {
```
pashakondratyev, line 14: We also would allow implicitly instantiating the `Ontology` and `Objects class` should we continue allowing that?

examples/one_dot_one/src/generatedNoCheck/Ontology.ts
mfedderly: Probably? Safer to move it than to not.

examples/one_dot_one/src/generatedNoCheck/Ontology.ts
mfedderly: The only thing I can think of is if someone wanted to have used a different client (or somehow a different stack?). That seems somewhat unlikely to be used?

mfedderly (approved):

examples/one_dot_one/src/generatedNoCheck/Ontology.ts
ericanderson: Having this here to see the effects of changes makes reviewing so nice.

PR #96, opened 2023-11-17
Remove const type parameters and fix build

mfedderly:


examples/basic/cli/src/examples/fetchAggregationForEmployeesGrouped.ts
```diff
@@ -23,16 +23,18 @@ import type { OntologyType } from "../OntologyType";
...
+    {
+      select: {
+        locationCity: "approximateDistinct",
+        locationName: "approximateDistinct",
+        employeeNumber: ["avg", "max", "min"],
+      },
+      groupBy: {
+        locationType: "exact",
+      },
+    } as const,
```
pashakondratyev, line 36: A bit concerned about this for 2.0 from a dev-ex perspective :(

mfedderly:
Moved the 2.0 client out of @osdk/api which alleviates this problem

PR #100, opened 2023-11-20
Typescript 4.9 support for v1.1 code

mfedderly:
- Change examples/one_dot_one to use ts4.9
- Fix errors coming from api, generator, and legacy-client related to the old compiler

### DIFF
--- a/.monorepolint.config.mjs
+++ b/.monorepolint.config.mjs
@@ -52,10 +52,28 @@ function standardPackageRules(shared) {
   return [
     standardTsconfig({
       ...shared,
+      excludePackages: [...shared.excludePackages, "@osdk/examples.basic.**"],
       options: {
         file: "tsconfig.json",
         template: {
-          extends: "tsconfig/base",
+          extends: "../../monorepo/tsconfig/tsconfig.base.json",
+
+          compilerOptions: {
+            rootDir: "src",
+            outDir: "build/types",
+            composite: true,
+          },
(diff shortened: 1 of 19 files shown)


.monorepolint.config.mjs
```diff
@@ -52,10 +52,28 @@ function standardPackageRules(shared) {
...
+          compilerOptions: {
+            rootDir: "src",
+            outDir: "build/types",
+            composite: true,
+          },
+          include: ["./src/**/*", ".eslintrc.cjs"],
+        },
+      },
+    }),
+    standardTsconfig({
```
mfedderly, line 70: @ericanderson is there a way to make this less duplicitive?

ericanderson: Either a function for reuse of the stnadardTsConfig. You can also give a generator function which passes you some details that might make it more programatic.

ericanderson: Also this is so sad. The old includes were so much better!

examples/one_dot_one/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "devDependencies": {
     "@types/node": "^18.0.0",
-    "typescript": "^5.2.2"
+    "typescript": "^4.9.0"
```
ericanderson, line 36: We should probably use MRL to make sure we don't break the invariant of certain packages being 4x

ericanderson (approved):

PR #106, opened 2023-11-27
Adding group by for array property types

pashakondratyev:


packages/legacy-client/src/client/baseTypes/ActionType.ts
```diff
@@ -120,12 +125,15 @@ export type ActionResponseFromOptions<
 
 export const ActionResponse = {
   of: <
-    TAddedObjects extends OntologyObject,
-    TModifiedObjects extends OntologyObject,
+    TAddedObjects extends OntologyObject<string, NonNullable<ParameterValue>>,
```
pashakondratyev, line 128: This was getting compiled down to something strange,
`OntologyObject<string, NonNullable<import("./ParameterValue")>>`, before which seemed off to me, and also messed with the generated .d.ts single file

PR #108, opened 2023-11-27
Add additional public API

mfedderly:
Some of the consumers may need these types due to weak type inference in places


pashakondratyev (approved):

mfedderly:
Not required due to https://github.com/ericanderson/osdk-ts/pull/109

PR #119, opened 2023-11-28
Recreate the objects-api types that support the ObjectSet consumers

mfedderly:


packages/legacy-client/src/index.ts
```diff
@@ -71,11 +72,15 @@ export type {
...
   GetObjectError,
+  GroupBySelections,
   ListLinkedObjectsError,
   ListObjectsError,
   LoadObjectSetError,
   MultiLink,
+  MultipleAggregateSelection,
+  ObjectTypeFilter,
   OntologyObject,
+  OrderBy,
```
mfedderly, line 83: ^ do we want to change any of these names before we make them public?

packages/legacy-client/src/client/objectSets/filters/Filters.ts
```diff
@@ -28,14 +27,6 @@ import type { GeoPointFilter } from "./GeoPointFilter";
...
 import type { StringFilter } from "./StringFilter";
 
-export type ObjectTypeFilterFunction<T extends ObjectTypeProperties> = (
-  objectType: T,
-) => WhereClause;
-
-export type ObjectTypeOrderByFunction<T extends ObjectTypeProperties> = (
-  objectType: T,
-) => OrderByClause;
-
```
mfedderly, line 38: This entire file might be unused, I have to look into it more.

pashakondratyev (approved):

PR #128, opened 2023-11-30
Backcompat shims for baseTypes/geoshapes directory

mfedderly:
First push doesn't have any of the autogenerated code checked in to test the build system
- [x] fails with changes in existing files not checked in
- [x] fails with changes in new files that got added

### DIFF
--- a/packages/generator/src/v1.1/generateObjectsInterfaceSupportFiles.test.ts
+++ b/packages/generator/src/v1.1/generateObjectsInterfaceSupportFiles.test.ts
@@ -34,28 +34,28 @@ describe(generateObjectsInterfaceSupportFiles, () => {
 
     expect(helper.getFiles()[`${BASE_PATH}/Todo.ts`]).toMatchInlineSnapshot(`
       "import { ObjectSetAggregateArg, ObjectSetFilterArg, ObjectSetGroupByArg, ObjectSetMultipleAggregateArg, ObjectSetOrderByArg } from \\"@osdk/legacy-client\\";
-      import { Todo } from \\"../Todo\\";
+      import { Todo as OG_Todo } from \\"../Todo\\";
 
       /** @deprecated Use Todo from ontology/objects instead */
-      export type Todo = Todo;
+      export type Todo = OG_Todo;
       /** @deprecated Use ObjectSetFilterArg<Todo> instead */
-      export type TodoFilter = ObjectSetFilterArg<Todo>;
+      export type TodoFilter = ObjectSetFilterArg<OG_Todo>;
       /** @deprecated Use ObjectSetOrderByArg<Todo> instead */
-      export type TodoOrderBy = ObjectSetOrderByArg<Todo>;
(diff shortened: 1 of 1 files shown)


packages/generator/src/v1.1/backcompat/internal-foundry-ontology-runtime-dist/baseTypes/generateGeoshapesDir.ts
```diff
@@ -0,0 +1,145 @@
...
+        "ofCentimeters",
+        "ofMeters",
+        "ofKilometers",
+        "ofInches",
+        "ofFeet",
+        "ofYards",
+        "ofMiles",
+        "ofNauticalMiles",
+      ]),
+    ),
```
mfedderly, line 68: This actually technically did not exist before

packages/legacy-client/src/client/baseTypes/geoshapes/index.ts
```diff
@@ -14,7 +14,11 @@
  * limitations under the License.
  */
 
+// eslint seems to incorrectly assume that `export * as Distance from "./DistanceMethods"` in Distance.ts exports the individual methods by name
+/* eslint-disable import/export */
 export * from "./Distance";
+export * from "./DistanceMethods";
+/* eslint-enable import/export */
```
mfedderly, line 21: ^ this is a weird eslint thing, we should probably file a bug report here at some point

packages/legacy-client/src/index.ts
```diff
@@ -141,6 +170,7 @@ export type {
   InvalidSortOrder,
   InvalidSortType,
   InvalidUserId,
+  LinearRing,
```
mfedderly, line 173: This entire set of additional APIs feels really bad that we're exporting them, especially without renames.

examples/todoapp/src/generatedNoCheck/internal/@foundry/ontology-runtime/dist/baseTypes/geoshapes/DistanceMethods.ts
```diff
@@ -0,0 +1,38 @@
+import {
+  ofCentimeters as OG_ofCentimeters,
```
pashakondratyev, line 2: We probably don't need to export these and can just export distance, since they're a bit strange by themselves (unless there's a reason for this)

ericanderson: If the treehsaking is a problem we dont need to make it better for 1.1

mfedderly:
```
.error TS2742: The inferred type of 'Distance' cannot be named without a reference to '$path'. This is likely not portable. A type annotation is necessary.
```

pashakondratyev (approved):

PR #131, opened 2023-12-01
Generating timeseries stubs for BackCompat

pashakondratyev:


examples/one_dot_one/src/generatedNoCheck/internal/@foundry/ontology-runtime/dist/aggregations/ActionType.ts
```diff
@@ -0,0 +1,2 @@
+
+
```
mfedderly, line 2: This is empty, probably a file that got generated but then never populated?

packages/generator/src/v1.1/backcompat/internal-foundry-ontology-runtime-dist/generateFiltersDir.ts
```diff
@@ -0,0 +1,170 @@
...
+    await formatTs(
+      `export * from "./ArrayFilter";
+      export * from "./AttachmentFilter";
+      export * from "./BooleanFilter";
+      export * from "./DateTimeFilters";
+      export * from "./Filters";
+      export * from "./GeoPointFilter";
+      export * from "./GeoShapeFilter";
+      export * from "./NumericFilter";
+      export * from "./Op";
```
mfedderly, line 41: You might have to re-const the Op stuff to make it useable.

pashakondratyev: It's re-exported as a const in `./Op.ts`, and that seems to work

mfedderly (approved):

PR #140, opened 2023-12-06
Start of interfaces

ericanderson:


packages/api/src/ontology/Definition.ts
```diff
@@ -27,26 +31,51 @@ export type ObjectInfoFrom<
...
   O extends OntologyDefinition<any>,
   K extends ObjectTypesFrom<O>,
-  P extends PropertyKeysFrom<O, K>,
-> = PropertyDefinitionsFrom<O, K>[P];
+  P extends ObjectPropertyKeysFrom<O, K>,
+> = ObjectPropertyDefinitionsFrom<O, K>[P];
+
+export type InterfaceInfoFrom<
+  O extends OntologyDefinition<any>,
+  K extends InterfaceNamesFrom<O>,
```
mfedderly, line 52: Shall we start using `I` instead of `K` which would normally be the objectTypeApi?

packages/client/src/objectSet/ObjectSet.ts
```diff
@@ -15,41 +15,44 @@
...
-    L
-  >; // TODO accept args?
-};
+  K extends ObjectTypesFrom<O> | InterfaceNamesFrom<O>,
+> = BaseObjectSet<O, K>;
 
 export interface BaseObjectSet<
   O extends OntologyDefinition<any>,
-  K extends ObjectTypesFrom<O>,
+  K extends ObjectTypesFrom<O> | InterfaceNamesFrom<O>,
```
mfedderly, line 39: Interfaces and Objects are guaranteed to have unique names within the union right?

mfedderly (approved):

PR #143, opened 2023-12-06
Renames

ericanderson:
- Rename ThinClient to ClientContext
- createThinClient -> createClientContext
- variables named thinClient -> clientCtx
- ValidPropertyTypes -> WirePropertyTypes
- Migrate some 2.0 types out of api


packages/api/src/index.ts
```diff
@@ -36,9 +32,8 @@ export type {
   ObjectPropertyKeysFrom,
   ObjectTypesFrom as ObjectTypesFrom,
   OntologyDefinition,
-  OsdkObjectPropertyType,
   PropertyDefinition,
-  ValidPropertyTypes as ValidPropertyTypes,
+  WirePropertyTypes as WirePropertyTypes,
```
mfedderly, line 36: Run through and clean up the duplicative `as` blocks

packages/client/package.json
```diff
@@ -52,5 +53,6 @@
   ],
   "main": "./build/js/index.js",
   "module": "./build/js/index.mjs",
-  "types": "./build/types/index.d.ts"
+  "types": "./build/types/index.d.ts",
+  "type": "module"
```
mfedderly, line 57: We want to not do this, right?

packages/shared.net/src/createClientContext.ts
```diff
@@ -14,25 +14,26 @@
...
-import type { ThinClient } from "./ThinClient";
+} from "./util/index.js";
 
 /**
  * The goal of the thin client is to provide a way to tree shake as much as possible.
  */
-
-export function createThinClient<T extends OntologyDefinition<any>>(
+export function createClientContext<
+  T extends { metadata: { userAgent: string } },
```
mfedderly, line 29: Its weird to still have the `T` arg be called `ontology`

packages/api/src/junk.test.ts
```diff
@@ -14,5 +14,10 @@
...
 
-export { createFetchHeaderMutator } from "../util/createFetchHeaderMutator.js";
-export { createRetryingFetch } from "../util/createRetryingFetch.js";
+import { describe, expect, it } from "vitest";
+
+describe("anything", () => {
+  it("does", async () => {
+    expect(1).toBe(1);
+  });
+});
```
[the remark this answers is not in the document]
ericanderson: without this the build gets mad you have no tests.

packages/client/package.json
ericanderson: I'm okay with it in `client` and it seems to work fine in my legacy check.

mfedderly: This is going to make it hard for anyone downstream to consume the package, just like we hit with conjure-lite

PR #152, opened 2023-12-07
Subscription devex cleanup

mfedderly:
- subscription is no longer async which was a ton of hassle for little gain
- watcher -> listener
- inline ObjectSetListener definitions so we can doc comment them


packages/client/src/objectSet/ObjectSetListener.ts
```diff
@@ -17,23 +17,16 @@
...
 > = Partial<
   {
-    [E in keyof ObjectSetWatcherEvents<O, K>]:
-      ObjectSetWatcherEvents<O, K>[E] extends never ? () => void : (
-        data: ObjectSetWatcherEvents<O, K>[E],
-      ) => void;
+    /** a specific list of objects have changed */
+    change: (objects: Array<OsdkObject<K & string>>) => void;
+    /** the objectset has become outdated and should be re-fetched in its entirety */
+    refresh: () => void;
```
pashakondratyev, line 28: Rather than `refresh` should this be something like `doRefresh` or `outOfSync`, since `refresh` seems a bit vague (without reading the doc)

pashakondratyev, line 28: For future iterations, should we call this something else? `refresh` seems vague in terms of what it means `doRefresh` or `outOfSync`

pashakondratyev (approved):

PR #153, opened 2023-12-07
MVP Actions in v2

ericanderson:


examples/basic/sdk/src/generatedNoCheck/Ontology.ts
```diff
@@ -1,18 +1,25 @@
...
-    Employee: Employee,
-    ObjectTypeWithAllPropertyTypes: ObjectTypeWithAllPropertyTypes,
+    Todo: Objects.Todo,
+    Person: Objects.Person,
+    Employee: Objects.Employee,
+    ObjectTypeWithAllPropertyTypes: Objects.ObjectTypeWithAllPropertyTypes,
+  },
+  actions: {
+    actionTakesAllParameterTypes: Actions.actionTakesAllParameterTypes,
+    createTodo: Actions.createTodo,
```
mfedderly, line 16: This is a nice way of avoiding collisions.

packages/client/src/actions/executeAction.ts
```diff
@@ -0,0 +1,43 @@
...
+  actionApiName: A,
+  parameters?: Parameters<Actions<O>[A]>[0],
+  options?: Op,
+): Promise<void> {
+  const response = await applyActionV2(
+    createOpenApiRequest(client.stack, client.fetch),
+    client.ontology.metadata.ontologyApiName,
+    actionApiName as string,
+    {
+      parameters: parameters as any,
```
mfedderly, line 39: Can we actually get away without having to marshal all the types to the wire format like in v1.1?

packages/client/src/types/Actions.ts
```diff
@@ -0,0 +1,103 @@
...
+
+type ActionParameterKeysFrom<
+  O extends OntologyDefinition<any>,
+  K extends ActionKeysFrom<O>,
+> = keyof ActionDefinitionFrom<O, K>["parameters"];
+
+type ActionParameterDefinitionFrom<
+  O extends OntologyDefinition<any>,
+  K extends ActionKeysFrom<O>,
+  P extends keyof O["actions"][K]["parameters"],
```
mfedderly, line 46:
```suggestion
  P extends ActionParameterKeysFrom<O, K>,
```

packages/client/src/types/Actions.ts
```diff
@@ -0,0 +1,103 @@
...
+> = ActionDefinitionFrom<O, K>["parameters"][P];
+
+type ActionParameterTypeFrom<
+  O extends OntologyDefinition<any>,
+  K extends ActionKeysFrom<O>,
+  P extends keyof O["actions"][K]["parameters"],
```
mfedderly, line 52: `ActionParameterKeysFrom` again

packages/client/src/types/Actions.ts
```diff
@@ -0,0 +1,103 @@
...
+> = NullableProps<ActionDefinitionFrom<O, K>["parameters"]> extends never
+  ? OsdkActionParametersNotOptional<O, K>
+  : PartialByNotStrict<
+    OsdkActionParametersNotOptional<O, K>,
+    NullableProps<ActionDefinitionFrom<O, K>["parameters"]>
+  >;
+
+export type Actions<O extends OntologyDefinition<any>> = {
+  [K in ActionKeysFrom<O>]: (
+    args: NOOP<OsdkActionParameters<O, K>>,
```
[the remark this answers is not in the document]
mfedderly: We should probably also hand the no arguments case here as well. I'm assuming there's actions that don't actually have any parameters?

packages/client/src/actions/executeAction.ts
pashakondratyev: Yeah we will need to handle these, but I don't think we've decided on the what attachments, timestamps, and dates will look like yet entirely. Objects and arrays of Objects wont work either I believe, and we need a way to extract the Object Set definition here as well. All the other primitives should be fine

ericanderson: For most things I was just going to lean into raw types as long as possible. Good points on the edge cases

PR #157, opened 2023-12-11
Adding login flow with CLI

pashakondratyev:
Still need to store the token somewhere safe/sane so that we can re-use it, but this should unblock some flows

### DIFF
--- a/packages/cli/src/commands/auth/CommonAuthArgs.ts
+++ b/packages/cli/src/commands/auth/CommonAuthArgs.ts
@@ -14,27 +14,8 @@
  * limitations under the License.
  */
 
-import type * as yargs from "yargs";
 import type { CliCommonArgs } from "../../CliCommonArgs.js";
-import type { CommonOntologyArgs } from "./CommonOntologyArgs.js";
-import login from "./login/index.js";
 
-const site: yargs.CommandModule<CliCommonArgs, CommonOntologyArgs> = {
-  command: "ontology",
-  describe: "Manage code",
-  builder: (argv) => {
-    return argv
-      .options({
-        baseUrl: {
-          type: "string",
-          demandOption: true,
-        },
-      })
-      .command(login)
-      .demandCommand();
-  },
-  handler: async (args) => {
-  },
-};
-
-export default site;
+export interface CommonAuthArgs extends CliCommonArgs {
+  baseUrl: string;
+}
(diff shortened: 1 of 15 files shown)


packages/cli/src/commands/auth/CommonAuthArgs.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { CliCommonArgs } from "../../CliCommonArgs.js";
+
+export interface CommonAuthArgs extends CliCommonArgs {
+  baseUrl: string;
```
mfedderly, line 20: We should probably be more consistent here with baseUrl vs stack

packages/cli/src/commands/auth/login/loginFlow.ts
```diff
@@ -0,0 +1,161 @@
...
+    resolve = _res;
+  });
+
+  const server = createServer((req, res) => {
+    const query = parse(req.url!, true).query;
+    res.end("Authenticated");
+    resolve(query["code"] as string);
+  });
+
+  server.listen(port);
```
mfedderly, line 40: Probably want to give a clean error message if the port is already in use by the app's webpack or whatever

packages/cli/src/commands/auth/login/loginFlow.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  consola.success(`Successfully authenticated!`);
+  return token;
+}
+
+function generateRandomString() {
+  const array = crypto.getRandomValues(new Uint32Array(28));
+  return Array.from(array, dec => {
+    const decimalString = "0" + dec.toString(16);
+    return decimalString.substring(decimalString.length - 2);
```
mfedderly, line 82: Took me a sec to figure this out. Also is a length of 28 part of a spec somewhere?

```suggestion
    return dec.toString(16).padStart(2,"0");
```

mfedderly (approved):

packages/cli/src/commands/auth/CommonAuthArgs.ts
pashakondratyev: I borrowed this from the `upload site` command but agreed

PR #160, opened 2023-12-11
Fixing crypto usage

pashakondratyev:


packages/cli/src/commands/auth/login/loginFlow.ts
```diff
@@ -82,20 +85,31 @@ export default async function invokeLoginFlow(args: LoginArgs) {
...
+
   consola.success(`Successfully authenticated!`);
+  consola.log(token);
   return token;
 }
 
-function generateRandomString() {
-  const array = crypto.getRandomValues(new Uint32Array(28));
+function generateRandomString(length = 128) {
+  const characters =
```
pashakondratyev, line 103: We were using hex-decimal before which wasn't high enough entropy

mfedderly (approved):

PR #162, opened 2023-12-11
Allow direct connection to a stack for ontology generation

mfedderly:
Adds the ability to generate an ontology directly from a given stack, given its stack URL and a clientId.
Users can still generate from a json file if they prefer.

### DIFF
--- a/packages/cli/src/commands/typescript/generate/TypescriptGenerateArgs.ts
+++ b/packages/cli/src/commands/typescript/generate/TypescriptGenerateArgs.ts
@@ -16,7 +16,9 @@
 
 export interface TypescriptGenerateArgs {
   outDir: string;
-  ontologyPath: string;
+  ontologyPath?: string;
+  stack?: string;
+  clientId?: string;
   beta?: boolean;
   packageType: "commonjs" | "module";
 }
--- a/packages/cli/src/commands/typescript/generate/handleGenerate.mts
+++ b/packages/cli/src/commands/typescript/generate/handleGenerate.mts
@@ -15,17 +15,83 @@
  */
 
 import {
+  getOntologyFullMetadata,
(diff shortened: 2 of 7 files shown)


packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -180,6 +180,7 @@ export class ObjectSetListenerWebsocket<
     if (state) {
       const { subscriptionId, objectSet } = state;
       if (subscriptionId) {
+        state.subscriptionId = undefined;
```
mfedderly, line 183: Unrelated cleanup

packages/cli/src/commands/typescript/generate/handleGenerate.mts
```diff
@@ -14,18 +14,84 @@
...
+
+  const token = await invokeLoginFlow({
+    applicationId: clientId,
+    baseUrl: stack,
+    verbose: 0,
+  });
+  const { fetch } = createClientContext(
+    {
+      metadata: {
+        userAgent: "@osdk/cli/0.0.0 ()",
```
pashakondratyev, line 62: You can probably use `readCliVersionFromPackageJson` for now as well

pashakondratyev (approved):

packages/cli/src/commands/typescript/generate/TypescriptGenerateArgs.ts
```diff
@@ -16,7 +16,9 @@
 
 export interface TypescriptGenerateArgs {
   outDir: string;
-  ontologyPath: string;
+  ontologyPath?: string;
+  stack?: string;
```
pashakondratyev, line 20: We should also clean up the other two args to use stack

PR #168, opened 2023-12-15
Support interfaces in where clause/object set types

mfedderly:


packages/client/src/object/fetchPageOrThrow.ts
```diff
@@ -31,17 +33,23 @@ import { convertWireToOsdkObjects } from "./convertWireToOsdkObjects.js";
 
 export interface FetchPageOrThrowArgs<
   O extends OntologyDefinition<any>,
-  K extends ObjectTypeKeysFrom<O>,
-  L extends ObjectTypePropertyKeysFrom<O, K>,
+  K extends ObjectTypeKeysFrom<O> | InterfaceKeysFrom<O>,
```
pashakondratyev, line 36: Should we make a type like `ObjectOrInterfaceKeysFrom<O>` and `ObjectOrInterfaceTypeDefinition` or something along those lines, might help ensure we don't accidentally drop off half of it in the future?

pashakondratyev (approved):

PR #169, opened 2023-12-15
Basic orderBy syntax for discussion

mfedderly:
Supports this syntax:

```typescript
client.objects.Ticket
        .orderBy({ field: sortField, direction: "asc" }, { field: sortField2, direction: "desc" })
        .fetchPageOrThrow();
        
// OR (user can choose to do either, they're equivalent)

client.objects.Ticket
        .orderBy({ field: sortField, direction: "asc" })
        .orderBy({ field: sortField2, direction: "desc" })
        .fetchPageOrThrow();
```

### DIFF
--- a/packages/client/src/object/fetchPageOrThrow.ts
+++ b/packages/client/src/object/fetchPageOrThrow.ts
@@ -15,8 +15,10 @@
  */
 
 import type {
+  InterfaceDefinitionFrom,
   InterfaceKeysFrom,
   InterfacePropertyKeysFrom,
+  ObjectTypeDefinitionFrom,
   ObjectTypeKeysFrom,
   ObjectTypePropertyKeysFrom,
   OntologyDefinition,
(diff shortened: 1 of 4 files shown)


packages/client/src/object/fetchPageOrThrow.ts
```diff
@@ -54,6 +58,10 @@ export async function fetchPageOrThrow<
     type: "base",
     objectType,
   },
+  orderByArg?: OrderByTerm<
+    T extends InterfaceKeysFrom<O> ? InterfaceDefinitionFrom<O, T>
```
pashakondratyev, line 62: I think just missing a `ObjectOrInterfaceKeysFrom` here instead

packages/client/src/query/OrderBy.ts
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { InterfaceDefinition, ObjectTypeDefinition } from "@osdk/api";
+
+export type OrderByTerm<
+  T extends ObjectTypeDefinition<any, any> | InterfaceDefinition<any, any>,
```
pashakondratyev, line 20: Nit: and here

pashakondratyev (approved):

mfedderly:
This is perhaps not the correct syntax, and is also broken by https://github.com/ericanderson/osdk-ts/pull/200

PR #172, opened 2024-01-03
Move gateway codegen 

pashakondratyev:
This moves the code-gen that we have into the OSS repo to be used the same way we generate conjure bindings.
The code itself has not changed so all the src files under `/gateway-generator` do not need a thorough review, though some can use a refactor probably.


packages/gateway-generator/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@osdk/gateway-generator",
+  "version": "0.0.0",
+  "description": "",
+  "access": "public",
```
mfedderly, line 5: Can we get away with making this private and not publishing it?

mfedderly (approved):

PR #173, opened 2024-01-05
Block osdk generation into directories that have contents

mfedderly:
This prevents us from users generating their .tgz packages on their local machine and combining multiple runs worth of output into a single (broken) package.

### DIFF
--- a/packages/cli/src/commands/typescript/generate/handleGenerate.mts
+++ b/packages/cli/src/commands/typescript/generate/handleGenerate.mts
@@ -26,6 +26,7 @@ import {
 import { createClientContext, createOpenApiRequest } from "@osdk/shared.net";
 import { consola } from "consola";
 import * as fs from "node:fs";
+import type {} from "@osdk/generator../../../../../generator/src/MinimalFs.js";
 import invokeLoginFlow from "../../auth/login/loginFlow.js";
 import type { TypescriptGenerateArgs } from "./TypescriptGenerateArgs.js";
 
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/typescript/generate/handleGenerate.mts
```diff
@@ -26,6 +26,7 @@ import {
 import { createClientContext, createOpenApiRequest } from "@osdk/shared.net";
 import { consola } from "consola";
 import * as fs from "node:fs";
+import type {} from "@osdk/generator../../../../../generator/src/MinimalFs.js";
```
pashakondratyev, line 29: Did you mean to import the MinimalFs type? Also should we export the type from `@osdk/generator`

mfedderly: why didn't the autofix stuff get rid of this 🤦 . I was thinking that MinimalFs was a little weird to export as public API so I just duck-typed it below.

packages/generator/src/util/verifyOutdir.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Re-generating an ontology on top of itself is not supported and may leave old
+ * files around or refuse to update files with new contents in some cases.
+ */
+export async function verifyOutdir(outDir: string, fs: MinimalFs) {
+  // if outDir exists and is not empty, we cannot proceed with the generation
+  try {
+    const contents = await fs.readdir(outDir);
+    if (contents.length !== 0) {
+      throw new Error(
+        "outDir is not empty, please delete its contents and try again",
```
pashakondratyev, line 31: Should we log the directory as well? Also I'm not sure if we have consola in generator but might be good for formatting the error.

packages/cli/src/commands/typescript/generate/handleGenerate.mts
pashakondratyev: It can make sense to export since it's an argument to the generator, but no need to do it now

PR #175, opened 2024-01-08
Alphabetize property names in generated object definition files

mfedderly:
This should help us more quickly find if properties are included in an object or not when reading the generated source code.


mfedderly:
Should we do this with object/query/action names as well?

pashakondratyev (approved):

PR #180, opened 2024-01-09
ClientV2 Links

mfedderly:
Adds support for traversing links from objects.

```typescript
// for links without multiplicity we support the lookup directly
employee.$link.lead.get();

// for links with multiplicity you can either
lead.$link.peeps.fetchPageOrThrow() // fetch paged results
lead.$link.peeps.get(peepId) // or get a linked object by its PK directly
```

All three calls support being passed an options argument with a list of property names to get only a subset of the original object's properties.


packages/client/src/definitions/LinkDefinitions.ts
```diff
@@ -0,0 +1,85 @@
...
+export type OsdkObjectLinksEntry<
+  K extends ObjectTypeKeysFrom<O>,
+  O extends OntologyDefinition<any>,
+  L extends ObjectTypeLinkKeysFrom<O, K>,
+> = ObjectTypeLinkDefinitionFrom<O, K, L> extends { multiplicity: false } ? {
+    /** Load the linked object */
+    get: () => OsdkObjectFrom<
+      ObjectTypeLinkDefinitionFrom<O, K, L>["targetType"],
+      O
+    >;
```
mfedderly, line 49: needs to support `select` option

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -64,6 +122,13 @@ export function convertWireToOsdkObjects<
   for (const obj of objs) {
     Object.setPrototypeOf(obj, proto);
 
+    Object.defineProperty(obj, OriginClient, {
```
mfedderly, line 125: I considered pushing this into the proto but then it made the caching semantics of `createCachedOntologyTransform` worse. Right now it caches on `===` ObjectTypeDefinition, which actually caches pretty well across client recreations and such.

packages/client/src/object/fetchPageOrThrow.ts
```diff
@@ -69,13 +70,17 @@ export async function fetchPageOrThrow<
...
+    select: ((args?.select as string[] | undefined) ?? []), // FIXME?
   };
 
   if (args?.nextPageToken) {
     body.pageToken = args.nextPageToken;
   }
 
+  if (args?.pageSize != null) {
+    body.pageSize = args.pageSize;
+  }
```
mfedderly, line 82: Was there a reason this wasn't exposed before?

packages/client/src/object/getLinkedObjectByPkOrThrow.ts
```diff
@@ -0,0 +1,50 @@
...
+    linkTypeApiName,
+    linkedObjectPrimaryKey,
+    {
+      select: [],
+    },
+  );
+
+  return convertWireToOsdkObjects(
+    client,
+    client.ontology.objects[sourceApiName].links[linkTypeApiName].targetType,
```
mfedderly, line 47: Do we actually want to be passing in apiName here? Or just grab `__apiName` from each object we convert?

packages/client/src/object/getLinkedObjectOrThrow.ts
```diff
@@ -0,0 +1,47 @@
...
+    { pageSize: 1 },
+  );
+
+  if (result.data.length !== 1 || result.nextPageToken != null) {
+    throw new PalantirApiError(
+      `Expected a single result but got ${result.data.length} instead${
+        result.nextPageToken != null ? " with nextPageToken set" : ""
+      }`,
+    );
+  }
```
mfedderly, line 44: ^ is this how we want to handle non-conjure errors related to network requests?

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -36,6 +43,57 @@ function createPrototype<
...
+                }),
+            };
+          }
+        },
+      });
+    },
+    enumerable: false,
+    configurable: false,
+    writable: false,
+  });
```
mfedderly, line 95: ^ these methods need runtime code testing before I feel comfortable shipping it

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -36,6 +43,57 @@ function createPrototype<
   const objDef = ontology.objects[type];
   const proto = {};
 
+  Object.defineProperty(proto, "$links", {
+    get: function() {
+      const client = this[OriginClient] as ClientContext<O>;
```
ericanderson, line 48: Im happy to see this.

packages/client/src/object/fetchPageOrThrow.ts
ericanderson: Mistake on my part. Not intentional

packages/client/src/object/getLinkedObjectByPkOrThrow.ts
ericanderson: I am guessing that its equiv for now. In fact you could add an invariant check here until interface support that says that we expect them to be the same

packages/client/src/object/getLinkedObjectOrThrow.ts
ericanderson: I think it doesn't matter in this case. We aren't using any of the other fields. We aren't leveraging retry (nor would i expect doing it again to work). 

Using this vs Error doesnt change anything besides how someone consumes it, right?

packages/client/src/object/getLinkedObjectByPkOrThrow.ts
mfedderly: (OR you stop having to look up the API name entirely, which also matches the legacy-client behavior)

ericanderson (approved):

packages/client/src/definitions/LinkDefinitions.test.ts
```diff
@@ -0,0 +1,107 @@
...
+            Todos: {
+              get: <
+                A extends SelectArg<
+                  typeof MockOntology,
+                  "Todo",
+                  ObjectOrInterfacePropertyKeysFrom<
+                    typeof MockOntology,
+                    "Todo"
+                  >
+                >,
```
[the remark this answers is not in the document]
mfedderly: Yeah I had to build this whole test out piecemeal, if it breaks hopefully its easy to keep diffing it forward.

PR #194, opened 2024-01-12
Add script export tests

mfedderly:


.vscode/settings.json
```diff
@@ -11,6 +11,6 @@
     "editor.defaultFormatter": "dprint.dprint"
   },
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
ericanderson, line 14: whats this do

ericanderson (approved):

.vscode/settings.json
mfedderly: I think the vscode eslint extension just wants to change its config, it keeps automatically changing and I'm tired of having it only in my local.

PR #195, opened 2024-01-16
Refactor test packages

mfedderly:
Instead of @osdk/shared.test depending on @osdk/legacy-client, we invert the relationship so that @osdk/client can benefit from the mock ontology and related mock utilities.


ericanderson (approved):
This was surprisingly not so bad

PR #197, opened 2024-01-17
Add objects, actions, queries enumerability for backcompat

mfedderly:
When we moved to the proxy-based implementation with v1.1 it accidentally removed the ability for us to inspect what types that we had compiled in.

### DIFF
--- a/packages/legacy-client/src/client/actions/createActionProxy.ts
+++ b/packages/legacy-client/src/client/actions/createActionProxy.ts
@@ -23,7 +23,7 @@ import type { ActionArgs, Actions, WrappedActionReturnType } from "./actions";
 export function createActionProxy<
   O extends OntologyDefinition<any>,
 >(client: ClientContext<O>): Actions<O> {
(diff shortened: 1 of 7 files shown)


packages/legacy-client/src/client/actions/createActionProxy.ts
```diff
@@ -51,6 +51,19 @@ export function createActionProxy<
 
         return undefined;
       },
+      ownKeys(_target) {
+        return Object.keys(client.ontology.actions);
+      },
+      getOwnPropertyDescriptor(_target, p) {
```
ericanderson, line 57: Might be worth just constructing an object or a prototype in the future given we have to do so much work anyway.

ericanderson: On second thought, probably not. A proxy means we can potentially support calls that we don't know about at compile time.

ericanderson (approved):

PR #201, opened 2024-01-18
Add Attachments class to v2

mfedderly:
In order to be able to serialize Attachments for Actions, we need to be able to differentiate them when we convert to the wire spec. `{rid: string}` is probably not specific enough to safely create these payloads so we promote it to a full class instead.

legacy-client also added additional methods on the Attachments class to get metadata and then the actual data itself, but that is omitted from this PR.

This also improves our msw stubs to handle more of the `objectTypeWithAllPropertyTypes` objects.

### DIFF
--- a/packages/client/src/Definitions.ts
+++ b/packages/client/src/Definitions.ts
@@ -20,6 +20,7 @@ import type {
   OntologyDefinition,
   WirePropertyTypes,
 } from "@osdk/api";
+import type { Attachment } from "./object/Attachment.js";
 
 type MaybeArray<T extends { multiplicity?: boolean | undefined }, U> =
   T["multiplicity"] extends true ? Array<U> : U;
@@ -31,8 +32,12 @@ type MaybeNullable<T extends ObjectTypePropertyDefinition, U> =
 type Raw<T> = T extends Array<any> ? T[0] : T;
 type Converted<T> = T extends Array<any> ? T[1] : T;
 
+// certain data types must be converted at hydration time
+type HydrationConversions<T extends ObjectTypePropertyDefinition> =
+  T["type"] extends "attachment" ? Attachment : WirePropertyTypes[T["type"]];
+
 export type OsdkObjectPropertyType<T extends ObjectTypePropertyDefinition> =
-  MaybeNullable<T, MaybeArray<T, Converted<WirePropertyTypes[T["type"]]>>>;
+  MaybeNullable<T, MaybeArray<T, Converted<HydrationConversions<T>>>>;
 
 export type OsdkObjectRawPropertyType<T extends ObjectTypePropertyDefinition> =
   MaybeNullable<T, MaybeArray<T, Raw<WirePropertyTypes[T["type"]]>>>;
--- a/packages/client/src/object/Attachment.ts
+++ b/packages/client/src/object/Attachment.ts
@@ -14,22 +14,14 @@
  * limitations under the License.
  */
 
-import type { ListObjectsResponseV2 } from "@osdk/gateway/types";
+export interface Attachment {
+  rid: string;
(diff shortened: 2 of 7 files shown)


packages/client/src/Definitions.ts
```diff
@@ -31,8 +32,12 @@ type MaybeNullable<T extends ObjectTypePropertyDefinition, U> =
 type Raw<T> = T extends Array<any> ? T[0] : T;
 type Converted<T> = T extends Array<any> ? T[1] : T;
 
+// certain data types must be converted at hydration time
+type HydrationConversions<T extends ObjectTypePropertyDefinition> =
```
mfedderly, line 36: Open to better names here

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -48,6 +50,42 @@ function createPrototype<
...
+  T extends keyof O["objects"] & string,
+  O extends OntologyDefinition<any>,
+>(
+  ontology: O,
+  type: T,
+) {
+  const steps: Array<(o: Record<string, any>) => void> = [];
+
+  for (
+    const [key, value] of Object.entries(ontology.objects[type].properties)
```
mfedderly, line 64: There are actually not value types that require recursion (as long as we're OK with doing arrays right when they come in)

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -61,37 +99,17 @@ export function convertWireToOsdkObjects<
...
-    //     const value = obj[key] as string | undefined;
-    //     if (value !== undefined) {
-    //       obj[key] = Temporal.PlainDateTime.from(value);
-    //     }
-    //   }
-    // }
+  if (converter) {
+    for (const obj of objs) {
+      Object.setPrototypeOf(obj, proto);
+      converter(obj);
```
mfedderly, line 107: This is another step away from "everything is just a plain javascript object"

packages/client/src/object/Attachment.ts
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface Attachment {
+  rid: string;
+}
+
+export class Attachment implements Attachment {
```
ericanderson, line 21: Given these are the same name, what do we get by separating them?

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -48,6 +50,42 @@ function createPrototype<
   return proto;
 }
 
+// preprocess the ontology definition to more quickly apply object conversions when needed
```
ericanderson, line 53: To be nitty, is it faster to create a function that must be invoked that has to iterate over an array of steps to do more function invocations? Would it just be faster to iterate over the properties?

ericanderson (approved):

packages/client/src/object/convertWireToOsdkObjects.ts
mfedderly: If you have to examine every property for every object, that's potentially quite a bit slower than preprocessing the config -> transform and then reusing the transform over and over. Here's a vaguely representative jsbench:

image

packages/client/src/object/Attachment.ts
mfedderly: Its perhaps useless? In my mind I was buying something about assignability across multiple versions of the osdk by using the Attachment type carefully and only using the Attachment class when we're doing implementation stuff.

PR #204, opened 2024-01-20
Add where clause support for "contains" filter

haosharon:
- this is only a type change because the underlying implementation already supports generic { fooFilterType: value } style queries


mfedderly (approved):
This won't work if someone tries to use it to subscribe to updates for the ObjectSet, but that's a future concern.

PR #209, opened 2024-01-23
V2 Actions parity

mfedderly:
Changes:
- Make actions enumerable
- Remap incoming params correctly before sending them to the backend
- Actions throw the validation errors if they exist
- Actions return the edits if they were requested
- Add some quick tests to support all of the above

User-facing semantics:
- When calling an action, the options validateOnly and returnEdits cannot both be true at the same time
- The return type of calling an action depends on its arguments. validateOnly will cause it to return the validation result, returnEdits will cause it to return the edits, and otherwise it will return undefined.
- Calling an action for execution will throw an `ActionValidationError` with the validation on it, if the validation does not pass.
- We have aliases of the wire types for the edit data and the validation responses, in case we decide to change them in the future.


examples/todoapp/src/CreateTodoForm.tsx
```diff
@@ -3,11 +3,15 @@ import React, { useState } from "react";
 export default function CreateTodoForm({
   createTodo,
 }: {
-  createTodo: (title: string) => Promise<void>;
+  createTodo: (
+    title: string,
+    setError: (error: string | undefined) => void
```
ericanderson, line 8: This is a nice touch. I don't think I would have thought of it.

examples/todoapp/src/useTodos.tsx
```diff
@@ -126,17 +127,40 @@ export function useTodos() {
...
+                          `Todo must be between ${constraint.gte}-${constraint.lte} characters`
+                        );
+                      }
+                    }
+                  }
+                }
+              }
+            }
+
+            throw e;
```
ericanderson, line 162: I like that if youre lazy and dont catch the mutate fails.

packages/client/src/actions/applyAction.ts
```diff
@@ -33,34 +48,47 @@ export async function applyAction<
...
     },
   );
 
-  const bulkEdits = response.edits?.type == "largeScaleEdits"
-    ? response.edits.editedObjectTypes
-    : undefined;
-  const edits = response.edits?.type == "edits" ? response.edits : undefined;
-  const q = edits?.edits[0];
+  if (options?.validateOnly) {
+    return response.validation as ActionReturnTypeForOptions<Op>;
```
ericanderson, line 67: I like this too.

ericanderson (approved):
Ship it

mfedderly:
merged upstream

PR #211, opened 2024-01-24
Add create-osdk-app CLI

tzyl:


packages/create-osdk-app/src/cli.ts
```diff
@@ -0,0 +1,404 @@
...
+  } else {
+    consola.info(`Creating project directory`);
+    fs.mkdirSync(root, { recursive: true });
+  }
+
+  consola.info(`Copying files into project directory`);
+
+  const templateDir = path.resolve(
+    fileURLToPath(import.meta.url),
+    "../..",
```
tzyl, line 125: When run in vitest this looks for `create-osdk-app/src/cli.ts` -> `create-osdk-app/template-*`
When run in tsup bundle this looks for `create-osdk-app/build/js/index.mjs` -> `create-osdk-app/build/template-*`

PR #212, opened 2024-01-24
Fix @osdk/cli version lookup code

mfedderly:


packages/cli/src/readCliVersionFromPackageJson.ts
```diff
@@ -15,10 +15,14 @@
  */
 
 import * as fs from "node:fs";
+import { dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
```
ericanderson, line 21: uber uber nit: why keep a global variable you need once?

mfedderly: Because it was a comfy place to copypaste obviously.

PR #213, opened 2024-01-24
Fix built package types

mfedderly:
This PR changes the `build` directory of our generated packages. The main directories are now `build/esm` and `build/cjs`. Each directory contains its source and types.

In order to actually accomplish this, tsup actually now has 4 configuration entries that generate each combination of format and .d.ts or .js. We take extra care when generating the package.json files to align the exact file extensions with commonjs packages.

The large majority of the files changed in this PR are actually just updating the import syntax as the packages switch to the module type. The first 8 commits contain all of the bulk changes that are fairly mechanical.

This also tightens up the testing in a few ways: we check @arethetypeswrong/cli against all of the packages, and we make sure the CLIs themselves will at least run and print a help message.

Fixes the v1.1 generator not working with module type output

Updates the @osdk/gateway-generator code to generate module import syntax


ericanderson (approved):
Approving on skimming plus conversation we had about the changes.

PR #218, opened 2024-01-26
Include OSDK package and foundry-sdk-generator in user agent

tzyl:
Currently there is no information in the user agent identifying the OSDK package name/version or the OSDK generator version being used. We get information only about whether the package was generated using 1.1 or 2.0.

Old example in 1.1's generated ontology metadata:

```
userAgent: foundry-typescript-osdk/0.0.1
```

After this PR we include the OSDK package name/version and the `@osdk/foundry-sdk-generator` version.

New example in foundry-sdk-generator's src/generatedNoCheck ontology:

```
userAgent: 'foundry-typescript-osdk/0.0.1 @test-app/osdk/0.0.1 @osdk/foundry-sdk-generator/0.200.6',
```

I also removed the empty `()` parts of user agents as we don't actually add any additional information. I considered if we should pass this information through to `createClientContext` directly rather than through the ontology but I don't think that works well in 2.0 where the user is expected to call generic `createClient` with just the ontology metadata so the ontology metadata itself must contain this information.

### DIFF
--- a/packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.ts
+++ b/packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.ts
@@ -30,6 +30,7 @@ export async function generateClientSdkVersionTwoPointZero(
   fs: MinimalFs,
   outDir: string,
   packageType: "module" | "commonjs" = "commonjs",
+  additionalUserAgents: string[] = [],
 ) {
   await verifyOutdir(outDir, fs);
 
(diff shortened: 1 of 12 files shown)


packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.ts
```diff
@@ -30,6 +30,7 @@ export async function generateClientSdkVersionTwoPointZero(
   fs: MinimalFs,
   outDir: string,
   packageType: "module" | "commonjs" = "commonjs",
+  additionalUserAgents: string[] = [],
```
tzyl, line 33: This is unused for 2.0 as foundry-sdk-generator doesn't allow generating 2.0 yet

mfedderly:
Upstream https://github.com/palantir/osdk-ts/pull/20

PR #225, opened 2024-01-30
Make objects enumerable in 2.0

ssanjay1:


packages/client/src/ObjectSetCreator.test.ts
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MOCK_ORIGIN } from "@osdk/shared.test";
+import { beforeAll, describe, expect, it } from "vitest";
+import type { Client } from "./Client.js";
+import { createClient } from "./createClient.js";
+import { MockOntology } from "./util/test/mockOntology.js";
```
mfedderly, line 21: I'd like to get away from using the handwritten mock ontology and use the `msw` mocks instead, but lets not block this PR on that.

```typescript
import { Ontology as MockOntology } from "../generatedNoCheck/index.js";
```

mfedderly (approved):

